PR #3, opened 2023-07-23
Use /var/tmp instead of /tmp as tempdir

### LINKED ISSUE #2: [Tumbleweed] bug: fsfreeze fails when executing dracut
When this line gets executed in sdbootutil
run_command_live_output dracut --quiet --reproducible "$tmpdir/initrd" "$kernel_version"
the following error occurs:
dracut: Could not fsfreeze /tmp/sdboot.XXXXXXX

This is likely due to the fact that /tmp is mounted as tmpfs, which does not support fsfreeze:
fsfreeze -f /tmp/XXX
fsfreeze: /tmp/XXX: Freezing failed: The operation is not supported.

TobiPeterG:
fixes #2
Needs to be tested if it works on Microos

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -19,7 +19,7 @@ arg_arch=
 
 rollback=()
 
-tmpdir=$(mktemp -d -t sdboot.XXXXXX)
+tmpdir=$(mktemp -d /var/tmp/sdboot.XXXXXX)
 cleanup()
 {
 	local i


lnussel:
should be fixed in dracut to suppress errors from fsfreeze I guess

lnussel:
https://bugzilla.opensuse.org/show_bug.cgi?id=1213643

TobiPeterG:
Closing this as it has been fixed upstream

PR #5, opened 2023-07-23
Create boot menu entries when installed

### LINKED ISSUE #4: [Tumbleweed] bug: sdbootutil install doesn't create a boot entry in boot menu
When following this tutorial:
https://en.opensuse.org/Systemd-boot
and executing sdbootutil install, all files are copied to the correct location, however, no boot entry is created.
Manually booting the file is possible though.

TobiPeterG:
fixes #4
Tested on Tumbleweed, needs testing on MicroOS

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -669,7 +669,7 @@ install_sdboot()
 
 	mkdir -p "$boot_root/loader/entries"
 
-	esp_device=$(lsblk -o NAME,MOUNTPOINT -nl | grep '/boot/efi' | awk '{print $1}')
+	esp_device=$(readlink -f $(bootctl status | awk -F': ' '/ESP:/ {print $2; exit}'))
 
 	if [ -n "$esp_device" ]; then
 	    partition_number=${esp_device##*[!0-9]}
--- a/sdbootutil
+++ b/sdbootutil
@@ -676,8 +676,8 @@ install_sdboot()
 		done
 		install -D "$sdboot" "$boot_root$sdboot_dst/grub.efi"
 		echo "shim.efi,openSUSE Boot Manager" | iconv -f ascii -t ucs2 > "$boot_root/$sdboot_dst/boot.csv"
-        efibootmgr --create --label "openSUSE Secure Boot" --loader "$sdboot_dst/shim.efi"
-        efibootmgr --create --label "openSUSE Boot Manager" --loader "$sdboot_dst/grub.efi"
+        efibootmgr -q --create --label "openSUSE Secure Boot" --loader "$sdboot_dst/shim.efi"
+        efibootmgr -q --create --label "openSUSE Boot Manager" --loader "$sdboot_dst/grub.efi"
 		# boot entry point
 		for i in MokManager fallback; do
 			install -D "$shimdir/$i.efi" "$boot_root/EFI/BOOT/$i.efi"
@@ -688,7 +688,7 @@ install_sdboot()
 		install -D "$sdboot" "$boot_root$entry"
 		install -D "$sdboot" "$boot_root/EFI/BOOT/BOOT${firmware_arch^^}.EFI"
 		echo "${sdboot##*/},openSUSE Boot Manager" | iconv -f ascii -t ucs2 > "$boot_root/$sdboot_dst/boot.csv"
-        efibootmgr --create --label "openSUSE Boot Manager" --loader "$entry"
+        efibootmgr -q --create --label "openSUSE Boot Manager" --loader "$entry"
 	fi
 	mkdir -p "$boot_root/$entry_token"
 	mkdir -p "/etc/kernel"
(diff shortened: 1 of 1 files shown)


lnussel:
That would create duplicate entries if called again though, right?

lnussel:
it does. also needs -p and -d options to tell it which disk, blindly defaults to sda1 which is not necessarily correct

TobiPeterG:
> it does. also needs -p and -d options to tell it which disk, blindly defaults to sda1 which is not necessarily correct

I added checks to only create the entries when there is none with the name already existing and added disk and partition number as arguments.
The function fails when the ESP couldn't be detected.

sdbootutil
```diff
@@ -669,13 +669,30 @@ install_sdboot()
 
 	mkdir -p "$boot_root/loader/entries"
 
+	esp_device=$(lsblk -o NAME,MOUNTPOINT -nl | grep '/boot/efi' | awk '{print $1}')
```
lnussel, line 672: you don't need grep and awk, awk alone would do :-)
Anyway there's already `$arg_esp_path` parsed from bootctl. bootctl also lists the partition, something like `/dev/disk/by-partuuid/f0a6cf18-e241-b4ee-68c3-ecd933f49e82`. So should be possible to read that link. Then get the partition from sysfs like this: https://github.com/lnussel/addimageencryption/blob/master/addimageencryption#L234

sdbootutil
```diff
@@ -669,13 +669,30 @@ install_sdboot()
...
 			install -D "$shimdir/$i.efi" "$boot_root$sdboot_dst/$i.efi"
 		done
 		install -D "$sdboot" "$boot_root$sdboot_dst/grub.efi"
 		echo "shim.efi,openSUSE Boot Manager" | iconv -f ascii -t ucs2 > "$boot_root/$sdboot_dst/boot.csv"
+
+        # Create boot menu entries if they do not exist
+        efibootmgr | grep -q 'Boot.*openSUSE Secure Boot' || efibootmgr -q --create --disk $drive --part $partition_number --label "openSUSE Secure Boot" --loader "$sdboot_dst/shim.efi"
+
+        efibootmgr | grep -q 'Boot.*openSUSE Boot Manager' || efibootmgr -q --create --disk $drive --part $partition_number --label "openSUSE Boot Manager" --loader "$sdboot_dst/grub.efi"
+
```
lnussel, line 695: should be sufficient to point at shim

lnussel: also the path needs be the fat path, right? Ie `\EFI\systemd\shim.efi`
and don't forget to quote the variables. The shellcheck program is helpful for that.

TobiPeterG: the path should be right, as least when I tested it, it worked :sweat_smile: 
I removed the grub entry and quoted the variables

sdbootutil
TobiPeterG: Ah I see, bootctl status also lists the ESP. arg_esp_path links to /boot/efi right? (so the path ralative to root)
I now use bootctl's listing, is this solution now better? :)

lnussel: how about using `$arg_esp_path` directly?

Note just cutting off numbers to get rid of the partition is not sufficient. Think of `/dev/loop0p1`. You need to go through sysfs as pointed out in the link

sdbootutil
lnussel: systemd explicity turns around the slashes so there must be a reason for it. maybe some firmwares accept unix paths but that does not mean it's correct

TobiPeterG: I have done some testing, efibootmgr automatically converts forwards slashes to backwards slashes.
This is also stated in the arch wiki:
https://wiki.archlinux.org/title/Unified_Extensible_Firmware_Interface#efibootmgr
`Note: UEFI uses backward slash \ as path separator but efibootmgr automatically converts UNIX-style / path separators.`

sdbootutil
TobiPeterG: It should work now, I'm also handling the case where it points to an image, is this useful?

sdbootutil
```diff
@@ -718,13 +718,34 @@ install_sdboot()
 
 	mkdir -p "$boot_root/loader/entries"
 
+	if [ -d "$arg_esp_path" ] || [ -b "$arg_esp_path" ]; then
+	    if [ -b "$arg_esp_path" ]; then
+	        blkpart="$arg_esp_path"
```
lnussel, line 723: is there reason to believe that arg_esp_path may refer to a block device?

sdbootutil
```diff
@@ -718,13 +718,34 @@ install_sdboot()
...
+	    else
+	        mountpoint -q "$arg_esp_path" || { echo "$arg_esp_path is not a valid mountpoint"; return 1; }
+	        blkpart="$(findmnt -nvo SOURCE "$arg_esp_path")"
+	    fi
+	    [ -L "/sys/class/block/${blkpart##*/}" ] || { echo "$blkpart is not a partition"; return 1; }
```
lnussel, line 728: I think you can just use err() here to make the tool exit

lnussel: (and in the other error cases too)

lnussel:
merged, I'll adjust to my liking :)

PR #7, opened 2023-07-23
Add default loader config when installed

### LINKED ISSUE #6: [Tumbleweed] bug: sdbootutil install doesn't create a loader.conf file
When executing bootctl install, a loader.conf file is generated at ESP/loader/loader.conf
This is not the case with sdbootutil install.

TobiPeterG:
Closes #6 
Tested on Tumbleweed, needs testing on MicroOS

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -693,6 +693,10 @@ install_sdboot()
 	update_random_seed
 
 	[ -s "$boot_root/loader/entries.srel" ] || echo type1 > "$boot_root/loader/entries.srel"
+
+    if [[ ! -e $boot_root/loader/loader.conf ]]; then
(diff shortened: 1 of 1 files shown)


sdbootutil
```diff
@@ -693,6 +693,10 @@ install_sdboot()
 	update_random_seed
 
 	[ -s "$boot_root/loader/entries.srel" ] || echo type1 > "$boot_root/loader/entries.srel"
+
+    if [[ ! -e $boot_root/loader/loader.conf ]]; then
+        echo -e "#timeout 3\n#console-mode keep\n" > $boot_root/loader/loader.conf
+    fi
 }
```
lnussel, line 700: [ -e "$boot_root/loader/loader.conf" ] || echo -e "#timeout 3\n#console-mode keep\n" > "$boot_root/loader/loader.conf"

PR #13, opened 2023-08-28
Do not add rootflags if is not transactional

aplanas:


aplanas:
The fix should be do a `sdbootutil set-default-snapshot 1` early before the sdbootutil kernel installation

PR #15, opened 2023-09-12
More descriptive bootmenu entries

TobiPeterG:
This creates more descriptive boot menu entries using a similar naming scheme as GRUB's implementation.
Tested only on openSUSE Tumbleweed, but should work on MicroOS as well

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -244,8 +244,10 @@ install_kernel()
 	calc_chksum "$src"
 	dst="/$entry_token/$kernel_version/linux-$chksum"
 
-	local snapshot="${subvol#@/.snapshots/}"
-	snapshot="${snapshot%/*}"
+    local s_dir="${subvol#@}"
+    s_dir="${s_dir%/*}"
+    local snap_path="${s_dir%/*}"
+	local snapshot="${s_dir#/.snapshots/}"
 
 	local initrd="${src%/*}/initrd"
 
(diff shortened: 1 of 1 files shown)


sdbootutil
```diff
@@ -289,15 +291,54 @@ install_kernel()
...
+
+	title="${important}${os_release_PRETTY_NAME} ${os_release_VERSION_ID} (${full_desc})"
+
 	# shellcheck disable=SC2154
 	sort_key="$os_release_ID"
+
 	entry_machine_id=
 	[ "$entry_token" = "$machine_id" ] && entry_machine_id="$machine_id"
 
+	if [ "$os_release_ID" == "opensuse-tumbleweed" ]; then
```
lnussel, line 333: [ "$(stat -f -c %T /etc)" = "overlayfs" ]

lnussel (changes requested):
I can't help it. I hate the xmllint calls. Maybe we need a `snapper info` command or a parameter to `snapper list` to only list one snapshot. Meanwhile
`sed -ne 's/.*<\(type\|date\|description\|key\|value\)>\(.*\)<.*/snapshot_\1="\2"/p' < info.xml ` should do

TobiPeterG:
> I can't help it. I hate the xmllint calls. Maybe we need a `snapper info` command or a parameter to `snapper list` to only list one snapshot. Meanwhile `sed -ne 's/.*<\(type\|date\|description\|key\|value\)>\(.*\)<.*/snapshot_\1="\2"/p' < info.xml ` should do

I found the snapper list command that's quite configurable with the --columns option, I use this instead now
Is the new code better? :)

sdbootutil
TobiPeterG: I copied the is_transactional() function from you snapper plugin, if it's correct now, please resolve :)

lnussel: better

lnussel:
Forking this amount of processes is still ugly and inefficient. However, when using snapper list there already is an `update_snapper()` function. And whith `--disable-used-space` it seems to be fast enough. So could actually pick the info from there by means of jq and a while loop as found in several places in the script.

I think it would also make sense to move mangling the title to a separate function. Something like
is_transactional || adjust_title_and_sorting

TobiPeterG:
> Forking this amount of processes is still ugly and inefficient. However, when using snapper list there already is an `update_snapper()` function. And whith `--disable-used-space` it seems to be fast enough. So could actually pick the info from there by means of jq and a while loop as found in several places in the script.
> 
> I think it would also make sense to move mangling the title to a separate function. Something like is_transactional || adjust_title_and_sorting

I oriented on how you read json files, now it's just one call.
I reused and expanded your update_snapper function, please check if you're happy with all changes :)

lnussel:
check https://github.com/openSUSE/sdbootutil/pull/17

lnussel:
merged via #17

PR #16, opened 2023-09-12
Draft: Add "normal" boot entries for Tumbleweed

TobiPeterG:
This adds regular boot entries for tumbleweed,
Tested on Tumbleweed
This feature would also work on MicroOS, but AFAIK it's not desired currently.

This PR is based on #15


TobiPeterG:
This currently works by creating new entries when a kernel is added (when it's the default subvolume) and remove them at rollback, copy the snapper entries and replace their values to create "new" normal entries.
The overhead of re-creating these normal entries for entries that could have been kept should be irrelevant in daily use.

lnussel:
not convinced yet :)

PR #17, opened 2023-09-28
Refactor boot descriptive entries

lnussel:


lnussel:
somehow freaks out systemd-boot. it displays alll entries as duplicates for some reason

TobiPeterG:
I think I see some issues in your code.
- you call  the set_snapper_title_and_sortkey "$snapshot" function, but I can't see where/how you use the values.
- your version still has the kernel_version in it, even though it's included in the title when the set_snapper_title_and_sortkey function is used. I added it to the title as the version string is in most cases not shown as the date/time differs
- You removed the full_desc attribute that might be set and relevant. It's shown in GRUB.

Maybe one of those explains why they are shown as duplicate entries?
I'll try to replicate the issue

lnussel:
> I think I see some issues in your code.
> 
>     * your version still has the kernel_version in it, even though it's included in the title when the set_snapper_title_and_sortkey function is used. I added it to the title as the version string is in most cases not shown as the date/time differs

Not shown but used for sorting. So I think we need it to get a stable order in case multiple kernels are installed, no?
 
>     * You removed the full_desc attribute that might be set and relevant. It's shown in GRUB.

Is it actually set by anything?

TobiPeterG:
> > I think I see some issues in your code.
> > ```
> > * your version still has the kernel_version in it, even though it's included in the title when the set_snapper_title_and_sortkey function is used. I added it to the title as the version string is in most cases not shown as the date/time differs
> > ```
> 
> Not shown but used for sorting. So I think we need it to get a stable order in case multiple kernels are installed, no?
> 
But when the kernel version is in the title, wouldn't it get sorted by kernel version when the rest of the title is the same anyway?

> > ```
> > * You removed the full_desc attribute that might be set and relevant. It's shown in GRUB.
> > ```
> 
> Is it actually set by anything?

That's what I don't know. The code comment said it was for user comments.

lnussel:
> > > I think I see some issues in your code.
> > > ```
> > > * your version still has the kernel_version in it, even though it's included in the title when the set_snapper_title_and_sortkey function is used. I added it to the title as the version string is in most cases not shown as the date/time differs
> > > ```
> > 
> > 
> > Not shown but used for sorting. So I think we need it to get a stable order in case multiple kernels are installed, no?
> 
> But when the kernel version is in the title, wouldn't it get sorted by kernel version when the rest of the title is the same anyway?

it's not sorted by title but by id AFAICS. So indeed for sorting it would not be relevant. Nevertheless since for rw volumes I have a short title now the kernel version seems necessary to display. Looks like this now:

sdbootmenu

Looking forward to see how it looks with your menu patch. I'd leave it like that for now.

TobiPeterG:
> > > > I think I see some issues in your code.
> > > > ```
> > > > * your version still has the kernel_version in it, even though it's included in the title when the set_snapper_title_and_sortkey function is used. I added it to the title as the version string is in most cases not shown as the date/time differs
> > > > ```
> > > 
> > > 
> > > Not shown but used for sorting. So I think we need it to get a stable order in case multiple kernels are installed, no?
> > 
> > 
> > But when the kernel version is in the title, wouldn't it get sorted by kernel version when the rest of the title is the same anyway?
> 
> it's not sorted by title but by id AFAICS. So indeed for sorting it would not be relevant. Nevertheless since for rw volumes I have a short title now the kernel version seems necessary to display. Looks like this now:
> 
> sdbootmenu
> 
> Looking forward to see how it looks with your menu patch. I'd leave it like that for now.

Alright, good to know. :)

I worked on the menu patch, fixed some issues I noticed in the last few days, but still haven't got any response so far. :(

The only think I'm unsure about is having the short title for the rw subvolumes as that might be confusing to users who don't know what these numbers mean (I'm thinking about my mom e.g. who wouldn't know what the difference was. But that's up to you ofc. :)

PR #23, opened 2023-11-20
Add systemd-boot-update-grub.service

aplanas:
Also add this service and systemd-boot-update.service in the multi-user.target.wants

Note: there is a very ugly part in the spec file. I need to enable the new service, and also another one provided by systemd. For doing that, instead of going for the preset path, I decided here to create an empty file, link it into the target and mark the empty service as `%ghost %{_unitdir}/systemd-boot-update.service`

### DIFF
--- a/sdbootutil.spec
+++ b/sdbootutil.spec
@@ -34,9 +34,9 @@ License:        MIT
 URL:            https://en.opensuse.org/openSUSE:Usr_merge
 Source:         %{name}-%{version}.tar
 Requires:       efibootmgr
-Requires:       systemd-boot
 Requires:       jq
 Requires:       sed
+Requires:       systemd-boot
 Supplements:    (systemd-boot and shim)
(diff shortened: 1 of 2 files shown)


lnussel:
I've actually implemented `sdbootutil update` for this. It's called from the snapper hook in transactional mode. I think for a rw system it should work to call `sdbootutil update` in `%post`.

lnussel:
or better a file trigger

aplanas:
Why not use the systemd provided service?

lnussel:
Doing it that way would mean one would always have to reboot twice to get the updated bootloader, no?

aplanas:
> Doing it that way would mean one would always have to reboot twice to get the updated bootloader, no?

To update the ESP bootloader no, only one reboot (that will be executed on the second reboot)

lnussel:
now in #28

PR #24, opened 2023-11-22
Generate new TPM2 predictions with pcr-oracle

aplanas:


lnussel:
pcr-oracle always updates all predictions, not just a specific one of eg the kernel that got just installed?

aplanas:
> pcr-oracle always updates all predictions, not just a specific one of eg the kernel that got just installed?

For now pcr-oracle predict the next update using the expected entry (via sort). This is WIP and will generate predictions for all the entries.

lnussel:
for our use the nicest solution would be to make pcr-oracle only operate on a specific entry, ie the one we prepare. Esp since sdbootutil stages all changes. Then copies everything, including potential rollback in case of eg out of space.

sdbootutil
```diff
@@ -21,6 +21,12 @@ arg_no_reuse_initrd=
 # for x in vmlinuz image vmlinux linux bzImage uImage Image zImage; do
 image=vmlinuz
 
+update_predictions=
+# TODO: The kernel is measured in PCR4 and initrd in PCR9.  When the
+# json gets stored in ESP, we need to replace PCR12 (cmdline) with
+# PCR9 (cmdline, initrd)
+pcrs="0,2,4,7,12"
```
lnussel, line 28: this list is also hardcoded in disk-encryption-tool. a case for a separate config?

sdbootutil
```diff
@@ -1007,16 +1030,21 @@ elif [ "$1" = "needs-update" ]; then
 	sdboot_needs_update "${2:-$root_snapshot}"
 elif [ "$1" = "update" ]; then
 	sdboot_needs_update "${2:-$root_snapshot}" && install_sdboot "${2:-$root_snapshot}"
+	update_predictions=1
```
lnussel: I'd move setting that variable into `install_sdboot`

sdbootutil
```diff
@@ -1007,16 +1030,21 @@ elif [ "$1" = "needs-update" ]; then
...
 elif [ "$1" = "force-update" ]; then
 	is_installed && install_sdboot "${2:-$root_snapshot}"
 elif [ "$1" = "add-kernel" ]; then
 	install_kernel "${3:-$root_subvol}" "$2"
+	update_predictions=1
```
lnussel: dito

lnussel (changes requested):

lnussel:
will be part of #28

PR #31, opened 2023-11-30
Predict all the entries

aplanas:
Also extract the PCR values form /etc/crypttab

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -22,13 +22,6 @@ arg_no_reuse_initrd=
 image=vmlinuz
 
 update_predictions=
-# TODO: The kernel is measured in PCR4 and initrd in PCR9.  When the
-# json gets stored in ESP. For how initr is named (with a hash) PCR12
-# (cmdline) cannot be used now until PCR9 (cmdline, initrd) can be
-# used via a dract module that extend initrd in memory.
-# TODO: Move it into a configuration file. Also replicated in
-# jeos-firstboot-enroll
-pcrs="0,2,4,7"
 
 rollback=()
 
@@ -363,8 +356,6 @@ install_kernel()
 		fi
 		log_info "generating new initrd"
 
-		# XXX: predictions need to be included in initrd at this point o_O
-		generate_tpm2_predictions
 		run_command_live_output dracut --quiet --reproducible "${dracut_args[@]}" "$tmpdir/initrd" "$kernel_version"
 	fi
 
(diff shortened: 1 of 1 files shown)


sdbootutil
```diff
@@ -962,24 +953,56 @@ set_default_snapshot()
...
-	tpm2_pcrread sha256:$pcrs > /etc/systemd/tpm2-pcr-signature.state || :
+	all_pcrs=$(get_pcrs)
+	[ -z "$all_pcrs" ] || rm -f /etc/systemd/tpm2-pcr-signature.json
+
+	# We make as many predictions as |all_pcrs| * |entries| to
+	# cover all the combinations.  pcr-oracle is smart to include
+	# the entry only one time, so we will not have duplications.
+	# This is a step for multi device configurations.
+	for pcrs in "$all_pcrs"; do
+		for entry in /boot/efi/loader/entries/*.conf; do
```
Vogtinator, line 988: Not necessarily for this PR, but does need some discussion: Which entries should be signed?

IMO only the current + latest should allow automatic unlock.

aplanas: Right. Once that all is put in place we need to revisit some of the decisions and designs. This will help a lot.

For this case I think that this decision is correct: the user should be allowed to boot in all the snapshots present in the system automatically.

For attacks that comes after exploiting CVEs present in old transactions, IMHO we should use a different mechanism: we need to re-enroll in the LUKS header a new TPM2 keyslot with a different RSA key, and all the affected snapshots deleted (with the boot entries associated with them)

But lets revisit this later.

Vogtinator: IMO no. It must be possible to keep old snapshots around, even ancient and known vulnerable ones.

More than the current + latest shoudn't be necessary for automated unlock, that's all that's needed to cover normal system use + (automated) rollback in case something went wrong.

Vogtinator: > For attacks that comes after exploiting CVEs present in old transactions, IMHO we should use a different mechanism: we need to re-enroll in the LUKS header a new TPM2 keyslot with a different RSA key, and all the affected snapshots deleted (with the boot entries associated with them)

Yes, to prevent downgrade attacks using old disk images it should be possible (and in certain cases done automatically) to effectively revoke access with the old keyslot data.

sdbootutil
```diff
@@ -962,24 +953,56 @@ set_default_snapshot()
...
+	local pcrs
+	# We can have multiple devices, each one of them with
+	# different PCRs
+	while read -r dev; do
+		pcrs="$(cryptsetup luksDump "$dev" | grep -Po 'tpm2-pubkey-pcrs:\s+\K.*')"
+		if [ -n "$pcrs" ]; then
+			pcrs="${pcrs//+/,}"
+			echo "$pcrs"
+		fi
+	done <<<$(blkid -t TYPE=crypto_LUKS -o device)
```
lnussel, line 967: That code would just take the pcrs from the first luks device, potentially unrelated to the current system. We already know the root device uuid though. So this should work:
`cryptsetup luksDump --dump-json-metadata "/dev/disk/by-uuid/$root_uuid" | jq '.tokens[]|select(.type == "systemd-tpm2")|."tpm2_pubkey_pcrs"|join(",")' -r`

sdbootutil
lnussel, line 988: `$boot_root/loader/entries`. ESP might be mounted somewhere else. However, now that pcr-oracle seems to support calculating predictions for a specific entry it seems to be logical to only process the entry sdbootutil actually just created. Updating all existing predictions all the time seems to be a surprising side effect.

lnussel (changes requested):

sdbootutil
aplanas: I did that on purpose. We can have multiple disks with different PCRs. The idea is that this function return multiple strings, each will be a sequence like `"a,b,c"`, that is later iterated in line 987

sdbootutil

aplanas: The issue is that I am not sure that we will know what is created and what is destroyed. IMHO for now this is more correct: create a prediction for all the entries (or a subset, as @Vogtinator suggest)

sdbootutil
aplanas: For clarity: it is not just taking the PCRs for the first LUKS device, it is taking the PCRs of all LUKS devices.

lnussel (approved):

PR #36, opened 2023-12-18
Do not fail if LUKS2 section is empty

aplanas:


sdbootutil
```diff
@@ -963,7 +963,7 @@ get_pcrs() {
 	# different PCRs
 	while read -r dev; do
 		pcrs=$(cryptsetup luksDump --dump-json-metadata "$dev" | jq -r "$jq_pcr")
-		[ -n "$pcrs" ] && echo "$pcrs"
```
lnussel, line 966: That's why one usually uses reverse logic
`[ -z "$pcrs" ] || echo "$pcrs"`

aplanas: Oh I see. Done!

PR #38, opened 2024-01-12
Bind mount /etc inside the snapshot

aplanas:
If MicroOS is used, the /etc/crypttab version inside the snapshot /etc can be a base that is actually modified via an overlay.  This is the case when during the enrollment we add the use of TPM2.

When dracut is called via --sysroot, it will use the unmodified version of /etc/crypttab, generating an initrd that will not contain the TPM2 libraries (/usr/lib64/cryptsetup/libcryptsetup-token-systemd-tpm2.so), failling the open of the LUKS2 device.

This commit bind mount the external current /etc inside the snapshot, making it visible for dracut to generate a current valid initrd.


lnussel:
We'd need to bind mount the overlay that belongs to the given snapshot though, right? Just using the current one would not work for arbitrary snapshots.

lnussel:
So far it's in fstab but should be changed IMO (https://github.com/openSUSE/transactional-update/issues/117)

lnussel:
@laenion we need a way to get a fully set up snapshot in order to generate arbitrary initrds from it. Short of https://github.com/openSUSE/transactional-update/issues/117 which would allow us to parse the info ourselves, can we get a function from tukit to either return the underlay dirs or mount an existing snapshot properly right away?
After all it can set up new ones which is a very similar operation, just with an additional subvol clone?

aplanas:
For completion. @laenion suggested this too:

`tukit --continue=2 open; tukit call 3 <your commmand>; cp bla somewhere_else; tukit abort 3`

aplanas:
@lnussel updated the PR. 

This one mount the overlays as read-only. It takes the upperdir and prefix it into the lowerdir list, so it can avoid to set the upper and work directories during the mount.  This seems to stabilize the overlay.

Seems that mounting `/var` is not required.

Some observations (for a MicroOS with two snapshots):

  * When running from snapshot 2 to generate the initrd of snapshot 1 (`sdbootutil add-all-kernels 1`) I can see this kind of error during the execution of `dracut`.  Today I can see this error also using the original `mount --bind`, that was not present last week. So I am not sure if it is related with this change or not.
```
find: warning: ‘-name’ matches against basenames only, but the given pattern contains a directory separator (‘/’), thus the expression will evaluate to false all the time.  Did you mean ‘-wholename’?
...
gzip: /var/tmp/dracut.LRh5HK/initramfs/.snapshots/1/snapshot/usr/share/kbd/keymaps/i386/qwerty/es.map.gz: No such file or directory
gzip: /var/tmp/dracut.LRh5HK/initramfs/.snapshots/1/snapshot/usr/share/kbd/keymaps/xkb/es.map.gz: No such file or directory
```

  * When removing the current kernel from the second snapshot (`sdbootutil remove-all-kernels`) I see this error if I recreated the entries for the first snapshot:
```
opensuse-microos-6.6.11-1-kvmsmall-2.conf is the default boot entry
opensuse-microos-6.6.11-1-kvmsmall-2.conf is the selected boot entry
Removed "/boot/efi/opensuse-microos/6.6.11-1-kvmsmall/linux-4190f9a56e931d4394b78c1917e9ff1e3f50c212"
Removed "/boot/efi/opensuse-microos/6.6.11-1-kvmsmall/initrd-58015452d169f1d55382cdafa3004b08e4152103"
Removed /boot/efi/loader/entries/opensuse-microos-6.6.11-1-kvmsmall-2.conf
Fatal: Unable to open file /tmp/efimnt.ZlWJLC//opensuse-microos/6.6.11-1-kvmsmall/linux-4190f9a56e931d4394b78c1917e9ff1e3f50c212: No such file or directory
Error: Failed to install TPM predictions for opensuse-microos-6.6.7-1-kvmsmall-1.conf
```

lnussel:
The code assumes there is only one lowerdir. AFAIK there could be many, even though t-u nowadays tries to fold them

aplanas:
No, multiple lowerdirs are separated by colon: `lowerdir=dir1:dir2:...`[1]. The current code is using this, removing the `sysroot` and adjusting the base lowerdir (the last of the list).

The code is based on how t-u does something like this: https://github.com/openSUSE/transactional-update/blob/master/sbin/transactional-update.in#L339

[1] https://docs.kernel.org/filesystems/overlayfs.html#directories

PR #40, opened 2024-01-26
Use systemd-pcrlock

aplanas:


sdbootutil
```diff
@@ -1037,6 +1113,28 @@ generate_tpm2_predictions()
...
 
+
+# TODO: Maybe share this code outside
+generate_tpm2_predictions()
+{
+	[ -e /etc/crypttab ] || return 0
+	grep -q "tpm2-device" /etc/crypttab || return 0
+	[ -e /etc/sysconfig/fde-tools ] || return 0
+
+	. /etc/sysconfig/fde-tools
```
aplanas, line 1124: Lets discuss this

sdbootutil
```diff
@@ -988,47 +988,123 @@ set_default_snapshot()
...
+	done < <(bootctl list --json=short | jq --raw-output '.[] | select(.type == "type1") | [(.initrd[0] | sub("/"; "\\"; "g")), .options] | join(" ")')
+
+	n=0
+	while read -r i; do
+		n=$((n+1))
+		run_log /usr/lib/systemd/systemd-pcrlock \
+		    lock-kernel-initrd \
+		    --pcrlock=/var/lib/pcrlock.d/720-kernel-initrd.pcrlock.d/initrd-"$n".pcrlock \
+		    /boot/efi/"$i"
+	done < <(bootctl list --json=short | jq --raw-output '.[] | select(.type == "type1").initrd[0]')
```
lnussel, line 1081: I think you want to call update_entries_for_this_system(), then filter on $entryfile

sdbootutil
```diff
@@ -988,47 +988,123 @@ set_default_snapshot()
...
+	done < <(bootctl list --json=short | jq --raw-output '.[] | select(.type == "type1").linux')
+
+	n=0
+	while read -r i; do
+		n=$((n+1))
+		echo "initrd=$i" > /tmp/cmdline
+		run_log /usr/lib/systemd/systemd-pcrlock \
+		    lock-kernel-cmdline \
+		    --pcrlock=/var/lib/pcrlock.d/710-kernel-cmdline.pcrlock.d/cmdline-"$n".pcrlock \
+		    /tmp/cmdline
```
lnussel, line 1060: $tmpdir/cmdline

sdbootutil
```diff
@@ -988,47 +988,123 @@ set_default_snapshot()
...
-	local jq_pcr='.tokens[]|select(.type == "systemd-tpm2")|."tpm2_pubkey_pcrs"|join(",")'
-	# We can have multiple devices, each one of them with
-	# different PCRs
-	while read -r dev; do
-		pcrs=$(cryptsetup luksDump --dump-json-metadata "$dev" | jq -r "$jq_pcr")
-		[ -z "$pcrs" ] || echo "$pcrs"
-	done <<<$(blkid -t TYPE=crypto_LUKS -o device)
+run_log()
+{
+	logger -t sdbootutil "$($@ 2>&1)"
```
lnussel, line 993: systemd-cat might be the better alternative

sdbootutil
aplanas: I am not sure. I guess you mean `update_entires()`, or `update_entries_for_subvol()`?. The second one seems unrelated, as I do not know the subvol parameter, and the first one seems that also cannot be reused.

What I want to do is get the all initrds path. For now I assume that the `len(initrd) == 1` per entry

sdbootutil
aplanas: that was my first version, but after reading some more `logger` was the better option. It is higher level, more standard and already use `systemd-cat` internally

sdbootutil

sdbootutil
aplanas: should be done

sdbootutil
lnussel: we do something highly systemd specific, so  "more standard" by whatever means does not count :) Anyway according to the docs systemd-cat doesn't require reading the output of the tool into a shell variable and then pass it as command line (which is of limited size). The other alternative of using logger would be
`"$@" 2>&1 | logger -t sdbootutil`.

aplanas: > we do something highly systemd specific

I am aware, and that is why the first version used systemd-cat. But seems that not many people recommend it when I looked for it, and are favoring logger. Even the systemd-cat manual page has a link to it.

> The other alternative of using logger would be

Yes, done

aplanas:
The new commit limit to 3 the entries processed for both models: pcr-oracle and pcrlock. Maybe we can remove this limitation from pcr-oracle, or find a better way to group by `rootflags=`

sdbootutil
```diff
@@ -988,55 +988,164 @@ set_default_snapshot()
...
+	    if ! pcr-oracle \
+		 --private-key /etc/systemd/tpm2-pcr-private-key.pem \
+		 --from eventlog \
+		 --output /etc/systemd/tpm2-pcr-signature.json \
+		 --target-platform=systemd \
+		 --boot-entry "${entry_id}" \
+		 sign "$pcrs"; then
+		err "Failed to install TPM predictions for ${entry_id}"
+	    fi
+	done < <(bootctl list --json=short | jq --raw-output '.[0:2] | .[] | select(.type == "type1") | .id')
```
lnussel, line 1119: running for all entries, even from other distros again?

aplanas: Ah sorry. I did not realize that there is a filter by distro. I was thinking that was an optimization to avoid calling bootctl several times.

Done.

sdbootutil
aplanas: There is no pcrlock file for the first enrollment. I implemented this as a first option and parsing this file was easy but that will imply that during the first enrollment the installer needs to provide a fake pcrlock file.

Also the code path to get the PCRs will be different from pcr-oracle case.

This solution at least fix both issues.

aplanas: I am using the same file name (and variable) used by fde-tools ... that is also very ugly

sdbootutil
```diff
@@ -988,55 +988,166 @@ set_default_snapshot()
...
-			     --target-platform=systemd \
-			     --boot-entry "$entry_id" \
-			     sign "$pcrs"; then
-				err "Failed to install TPM predictions for $entry_id"
-			fi
-		done
-	done
+	# Publish the assets in the ESP, so can be imported by
+	# dracut-pcr-signature
+	[ -e /var/lib/systemd/pcrlock.json ] && cp /var/lib/systemd/pcrlock.json /boot/efi/EFI/systemd
```
lnussel, line 1098: `"$boot_root$sdboot_dst"` missed that one in other bits too

aplanas: right, done

aplanas:
I changed the way the snapshots are grouped. Instead of getting three entries, it gets now all the entries for the last three snapshots.

For pcr-oracle this seems to be the correct way, but for pcrlock maybe we need to limit it more

sdbootutil
aplanas: done with the new method

sdbootutil
```diff
@@ -1107,53 +1106,205 @@ set_default_snapshot()
...
+# TODO: Maybe share this code outside
+generate_tpm2_predictions_pcr_oracle()
+{
+	local pcrs="$1"
+
+	[ -z "$pcrs" ] || rm -f /etc/systemd/tpm2-pcr-signature.json
+
+	# Get the numbers for the last three snapshots
+	[ -s "$snapperfile" ] || update_snapper
+	local last_snapshots="$(jq --raw-output '.root | map(.number) | reverse | .[0:3] | map(select(. > 0)) | join("|")' "$snapperfile")"
```
lnussel, line 1267: this won't work for rw systems as the current snapshot could be some random number in the middle

sdbootutil
```diff
@@ -1107,53 +1106,205 @@ set_default_snapshot()
...
+	update_entries_for_snapshot "(:?${last_snapshots})"
```
lnussel, line 1268: creative abuse :)

sdbootutil
```diff
@@ -1107,53 +1106,205 @@ set_default_snapshot()
...
+		    "$tmpdir/cmdline.utf16"
+		rm "$tmpdir/cmdline.utf16"
+	done < <(jq --raw-output '.[] | ([(.initrd[0] | sub("/"; "\\"; "g")), .options] | join(" ")), .initrd[0]' "$entryfile")
+
+	# Generate variation for 710-kernel-cmdline-initrd-entry
+	# component that contains the current cmdline and the current
+	# initrd, even if this will never be used again.  This is
+	# required because disk-encryption-tool generates a new initrd
+	# during the first boot, making the event log impossible to
+	# align for systemd-pcrlock
```
lnussel, line 1219: shouldn't this code be refactored into a function that can be reused by above code?

sdbootutil
```diff
@@ -1107,53 +1106,205 @@ set_default_snapshot()
...
+	if [ "$SDB_ADD_INITIAL_CMDLINE" = "1" ]; then
+		local initrd="$(cut -d' ' -f1 /proc/cmdline | cut -d'=' -f2 | tr '\\' '/')"
```
lnussel, line 1221: Did I mention that I hat `cut`? Can be done with pure bash actually

sdbootutil
```diff
@@ -1107,53 +1106,205 @@ set_default_snapshot()
...
+		   "$tmpdir/cmdline-0.pcrlock" \
+		   "$tmpdir/initrd-0.pcrlock" \
+		   > /var/lib/pcrlock.d/710-kernel-cmdline-initrd-entry.pcrlock.d/cmdline-initrd-0.pcrlock
+		rm "$tmpdir/cmdline"
+		rm "$tmpdir/cmdline-0.pcrlock"
+		rm "$tmpdir/initrd-0.pcrlock"
+
+		# 710-kernel-cmdline-boot-loader.pcrlock.d is not part
+		# of the pcrlock standards
+		echo -ne "$(cat /proc/cmdline)\0" > "$tmpdir/cmdline"
```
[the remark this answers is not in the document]
aplanas: done

lnussel:
merged via #45 and fixes from #46

PR #47, opened 2024-02-14
Unify PCR selection

aplanas:


lnussel:
the issue with this is that this might introduce inconsistencies between the settings enrolled in the luks header vs the ones in the policy. pcrlock does not have this issue as the pcr list is not stored in the luks header.

aplanas:
The inconsistency appears in both solutions: with signed policy the PCR list is in the LUKS header, and with NV policy it is stored in the pcrlock.json. The problem is that in the LUKS header we can have a different list of PCR per device, and with pcrlock.json we need to share the same set of PCRs.

The current implementation is exploiting this difference, as acknowledge that different headers can have different set of PCRs in one case but not in another, making a very hard problem to debug

The real solution is to link the pcrlock.json with a device, enabling multiple of them, but this requires a change in systemd.  Until this happens, we need to go to the minimal common denominator here and use a single source of truth, like it was in the original patch.

lnussel:
I value the consistency within one setup more important than across tools here.  So if the LUKS header lists PCRs differnt from what's configured in sysconfig it does not make sense to generate a policy, pretending it would work. So either we use the information from the LUKS header to generate the policy, or refuse generating a policy if there's a mismatch.

aplanas:
> refuse generating a policy if there's a mismatch

We can do this. But how to do that in the pcrlock case?

lnussel:
With pcrlock that issue just doesn't exist. After reading the code I also wonder why it has to be like this with signed policies. There doesn't seem to be a need to actually store the pcr selection in the LUKS header.  The `find_signature()` function which iterates over `tpm2-pcr-signature.json` does this:

```
        /* Searches for a signature blob in the specified JSON object. Search keys are PCR bank, PCR mask,    
         * public key, and policy digest. */
```

Should be fully sufficient to just omit the pcr mask for searching purposes and take it from the file itself.

lnussel:
as long as `/var/lib/systemd/pcrlock.json` is not deleted (eg `pcrlock remove-policy`) there is no inconsistency. Command line parameters introduced by our tooling does not influence it. Anyway the more I look at the code, the more I wonder whether it makes sense to store any of that stuff in the luks header at all.

aplanas:
Is not as simple as that. In pcr-oracle there are two elements (json and luks2 header), but in pcrlock there are three elements (json, luks2 header and the TPM2 NV index). When the prediction fails you can remove the TPM2 content with `tpm2_clear` and now you have the inconsistent state (that fix is the same for the pcr-oracle case: re-enroll).

 ---

wrt `tpm2_pubkey_pcrs` I think that there is a reason why it is required. When cryptsetup try to unlock a device it load this field to build the `pubkey_pcr_mask`, that is a parameter required by `tpm2_unseal`. 

Another way of thinking about this is that if you do not know the PCRs involved in a policy, how the TPM2 can later build the policy that should match with one signed?  The policy hash does not contain information about how it was built. This information is never stored in the TPM2 nor in any policy blob.  Is the user the one that knows how that was built, as it can be as complicated as you want.

In signed policy the information is in the luks2 header, in pcrlock it is in the json (as contains the pcr set and the hashes variant, so you can build the PolicyAND and PolicyOR sequence that will built the policy hash)

lnussel:
Digressing here. It seems to me the way things are handled is grown organically and maybe needs a step back to reconsider. We need external files anyway now so IIUC there is no need to store anything in the luks header at all besides the fact that a certain key slot may be unlocked by means of tpm. Whatever is in the token at the moment could be stored in those external json files just as well.

TobiPeterG:
Is this PR still relevant?

aplanas:
Yeah not sure ... The idea is that depending on what mechanism we use (pcr-oracle, pcrlock) we are looking for the expected PCRs in different places, and this seems ugly.

But after a while I realize that there is no ideal solution, at least no one super clear.

The reason is that if signed policy is used, then we cannot add or remove PCRs once they are in the LUKS2 header, but with authorized policy we can add or remove those PCRs without rewriting the LUKS2 (re-enroll)

Yeah ... lets close it : ))

Issue #51, opened 2024-02-28
Bug: On Tumbleweed, the VERSION_ID in the title of regular entries is not always correct


TobiPeterG:
On Tumbleweed, it might happen that a new kernel gets installed and the version of the tumbleweed snapshot changes (e.g. 20240220 to 20240222), but the old kernel is not removed yet. This leads to a situation where the boot entry of the old kernel won't get updated and the boot menu shows the wrong tumbleweed version for the old kernel.
This might be confusing as the entry showing the older version will actually also boot the newer snapshot.

Issue #49, opened 2024-02-28
Title of snapshot is sometimes mixed up


thkukuk:
I installed openSUSE MicroOS with systemd-boot. After some days I did run transactional-update dup, and the version of the product in the snapshot were mixed up:

First entry:
```
# Boot Loader Specification type#1 entry 
title     openSUSE MicroOS 20240226 
version   1@6.7.5-1-default
```
Second entry:
```
# Boot Loader Specification type#1 entry 
title     openSUSE MicroOS 20240218 
version   2@6.7.6-1-default
```
The dates are clearly switched...
When I created some more snapshots with different packages added, afterwards all snapshots had `openSUSE MicroOS 20240226` as title.

PR #52, opened 2024-02-28
Remove Tumbleweed version from regular entries

TobiPeterG:
This fixes an issue where tumbleweed gets updated to a new snapshot with new kernel, but the version in entries is not adapted. The version will still be visible in snapper entries, as it won't change there.
Fixes https://github.com/openSUSE/sdbootutil/issues/51

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -363,6 +363,12 @@ umount_etc()
 	umount "${snapshot_dir}/etc"
 }
 
+add_version_to_title()
+{
+	# TW pretty name does not include the version
+	[ -n "$os_release_VERSION" ] || title="$title $os_release_VERSION_ID"
+}
+
 install_kernel()
 {
 	local subvol="$1"
@@ -429,12 +435,13 @@ install_kernel()
 	fi
 
 	title="${os_release_PRETTY_NAME:-Linux $kernel_version}"
-	# TW pretty name does not include the version
-	[ -n "$os_release_VERSION" ] || title="$title $os_release_VERSION_ID"
 	# shellcheck disable=SC2154
 	sort_key="$os_release_ID"
 
-	if ! is_transactional  && subvol_is_ro "$subvol"; then
+	if is_transactional; then
+		add_version_to_title
+	elif subvol_is_ro "$subvol"; then
+		add_version_to_title
 		set_snapper_title_and_sortkey "$snapshot"
 	fi
 


lnussel:
AFAICT this would also remove the version from microos entries

TobiPeterG:
> AFAICT this would also remove the version from microos entries

Ah alright
The comment said "on TW" that's why I thought it only applied to Tumbleweed.
Will fix it :)

TobiPeterG:
> AFAICT this would also remove the version from microos entries

Should be fixed now :)

sdbootutil
```diff
@@ -429,13 +435,16 @@ install_kernel()
...
-	if ! is_transactional  && subvol_is_ro "$subvol"; then
-		set_snapper_title_and_sortkey "$snapshot"
+	if ! is_transactional; then
+		if subvol_is_ro "$subvol"; then
+			add_version_to_title
+			set_snapper_title_and_sortkey "$snapshot"
+		fi
+	else
+		add_version_to_title
 	fi
```
lnussel, line 448:
```
if is_transactional; then
  add_version_to_title
elif subvol_is_ro "$subvol"; then
  ...
fi
```
And while you are at it, there is also #49 ... :-)

lnussel (changes requested):

sdbootutil
TobiPeterG: refactored it :)

#49 is probably caused by something else. I think I know the issue:
for i in /etc/os-release /usr/lib/os-release; do
	[ -f "$i" ] || continue
	eval $(sed -ne '/^[A-Z_]\+=/s/^/os_release_/p' < "$i")
	break
done
looks for the os-release in the current snapshot, not the new one

TobiPeterG: similarly, we're reading the current machine-id
[ -s /etc/machine-id ] && read -r machine_id < /etc/machine-id

Is there any way it changes between snapshots?

lnussel: the one in etc is just compat. The canonical one should be in /usr. Machine-id could change I guess but is unexpected. Would mean it's a different instance of the same OS. If we make everything snapshot specific we could deploy arbitrary states though. Anyway, looks like a bigger change, you don't have to to do that in the scope here indeed.

TobiPeterG: Can this be merged then or are there still open items?

Issue #20, opened 2023-10-05
Support for automatic boot assessment


TobiPeterG:
A first start to assess whether a boot was successful or not might be
https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/

It seems to be enough to add the wanted attempts by a +$number to the file name and pull in the required services/targets (graphical, multi user, network) to the boot-complete.target

PR #53, opened 2024-02-28
Add basic boot assessment

TobiPeterG:
This adds 3 attempts to all boot entries. If all attempts fail, it will automatically get marked unbootable. This commit also adds logic to all parts that access the conf files to detect them, no matter the state.

However, as listed here: https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/ we still need to add the adequate services as Requires= to boot-complete.target. This way, network etc. can also be assessed.
Boot entries will be assessed until they boot successfully once.
Closes https://github.com/openSUSE/sdbootutil/issues/20

Please test the changes here

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -284,6 +284,47 @@ sedrootflags()
 	sed "${sed_arguments[@]}"
 }
 
+entry_filter=("cat")
+update_entries()
+{
+	[ -z "$1" ] || entry_filter=("$@")
+	bootctl list --json=short | "${entry_filter[@]}" > "$entryfile"
+}
+
+update_entries_for_subvol()
+{
+	local subvol="$1"
+	update_entries jq "[.[]|select(has(\"options\"))|select(.options|match(\"root=UUID=$root_uuid .*rootflags=subvol=$subvol\"))]"
+}
+
+update_entries_for_snapshot()
+{
+	local n="$1"
+	update_entries_for_subvol "${subvol_prefix}/.snapshots/$n/snapshot"
+}
+
+update_entries_for_this_system()
+{
+	update_entries jq "[.[]|select(has(\"options\"))|select(.options|match(\"root=UUID=$root_uuid\"))]"
+}
+
+find_conf_file() {
+	local kernel_version="${1:?}"
+	local snapshot="$2"
+	local id="$entry_token-$kernel_version${snapshot:+-$snapshot}.conf"
+
+	update_entries_for_snapshot "$snapshot"
+
+	while IFS= read -r path; do
+		if [ -f "$path" ]; then
+			echo "$path"
+			return 0
+		fi
+	done < <(jq -r --arg id "$id" '.[] | select(.id == $id) | .path' < "$entryfile")
+
+	return 1
+}
+
 remove_kernel()
 {
 	local snapshot="$1"
@@ -315,7 +356,7 @@ install_with_rollback()
 
 update_snapper()
 {
-    snapper --jsonout --no-dbus list --disable-used-space > "$snapperfile"
+	snapper --jsonout --no-dbus list --disable-used-space > "$snapperfile"
 }
 
 set_snapper_title_and_sortkey()
@@ -349,13 +390,26 @@ set_snapper_title_and_sortkey()
 reuse_initrd() {
 	local snapshot="$1"
 	local subvol="$2"
+	local kernel_version="${3:?}"
+	local conf
 
 	[ -z "$arg_no_reuse_initrd" ] || return 1
 
-	local conf="$boot_root/loader/entries/$entry_token-$kernel_version-$snapshot.conf"
-	if [ -e "$conf" ]; then
-		local k
-		local v
+	conf="$(find_conf_file "$kernel_version" "${snapshot}")"
+	local find_conf_status=$?
+
+	if [ $find_conf_status -ne 0 ]; then
+		# check if we can reuse the initrd from the parent
+		# to avoid expensive regeneration
+		detect_parent "$subvol"
+		if [ -n "$parent_subvol" ]; then
+			conf="$(find_conf_file "$kernel_version" "$parent_snapshot")"
+			find_conf_status=$?
+		fi
+	fi
+
+	if [ "$find_conf_status" -eq 0 ]; then
+		local k v
 		while read -r k v; do
 			[ "$k" = 'initrd' ] || continue
 			log_info "found existing initrd $v"
@@ -363,23 +417,6 @@ reuse_initrd() {
 		done < "$conf"
 		[ -z "$dstinitrd" ] || return 0
 	fi
-
-	# check if we can reuse the initrd from the parent
-	# to avoid expensive regeneration
-	detect_parent "$subvol"
-	local parent_conf="$boot_root/loader/entries/$entry_token-$kernel_version-$parent_snapshot.conf"
-	if [ -n "$parent_subvol" ] && [ -e "$parent_conf" ]; then
-		#subvol_is_ro "$parent_subvol" || err "Parent snapshot $parent_snapshot is not read-only, can't reuse initrd"
-		local k
-		local v
-		while read -r k v; do
-			[ "$k" = 'initrd' ] || continue
-			log_info "found parent initrd $v"
-			dstinitrd+=("$v")
-		done < "$parent_conf"
-		[ -z "$dstinitrd" ] || return 0
-	fi
-
 	return 1
 }
 
@@ -441,7 +478,7 @@ install_kernel()
 			((++i))
 		done
 		/usr/bin/mkmoduleinitrd "${subvol#"${subvol_prefix}"}" "$kernel_version" "$tmpdir/initrd-$i"
-	elif ! reuse_initrd "$snapshot" "$subvol"; then
+	elif ! reuse_initrd "$snapshot" "$subvol" "$kernel_version"; then
 		local snapshot_dir="/.snapshots/$snapshot/snapshot"
 		local dracut_args=()
 		dracut_args=('--force' '--tmpdir' '/var/tmp')
@@ -518,7 +555,16 @@ install_kernel()
 		done
 	fi
 	if [ -z "$failed" ]; then
-		loader_entry="$boot_root/loader/entries/$entry_token-$kernel_version${snapshot:+-$snapshot}.conf"
+		local tries
+		if [ -f /etc/kernel/tries ]; then
+			read -r tries < /etc/kernel/tries
+		fi
+
+		if ! [[ "$tries" =~ ^[0-9]+$ ]]; then
+			tries=
+		fi
+
+		loader_entry="$boot_root/loader/entries/$entry_token-$kernel_version${snapshot:+-$snapshot}${tries:++$tries}.conf"
 		install_with_rollback "$tmpdir/entry.conf" "$loader_entry" || failed="bootloader entry"
 		rm -f "$tmpdir/entry.conf"
 	fi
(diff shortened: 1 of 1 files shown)


sdbootutil
```diff
@@ -240,13 +240,29 @@ sedrootflags()
 	sed -e "s/[ \t]\+/ /g;s/\<\(BOOT_IMAGE\|initrd\)=[^ ]* \?//;s/\<root=[^ ]*/root=UUID=$root_uuid/;tr;:r;s,\<rootflags=subvol=[^ ]*,rootflags=subvol=$subvol,;tx; s,\$, rootflags=subvol=$subvol,;:x"
 }
 
+find_conf_file() {
+	local snapshot="${1:?}"
+
+	local file suffix
+	for file in "${conf_dir}/${entry_token}-${kernel_version}-${snapshot}"*.conf; do
```
lnussel, line 247: Hmm, you check whether it's easier to use jq for that? we have `update_entries_for_snapshot` and in the json output systemd does the parsing of the counter for us:

```
                "id" : "opensuse-tumbleweed-6.7.5-1-kvmsmall-1.conf",
                "path" : "/boot/efi/loader/entries/opensuse-tumbleweed-6.7.5-1-kvmsmall-1+3.conf",
```

lnussel (changes requested):

sdbootutil
TobiPeterG: Adapted the code, hope it's what you had in mind :)

sdbootutil
```diff
@@ -240,13 +240,56 @@ sedrootflags()
...
+find_conf_file() {
+	local snapshot="${1:?}"
+
+	if [ ! -s "$entryfile" ]; then
+		update_entries_for_snapshot "$snapshot"
+	fi
+
+	jq -r --arg snapshot "${snapshot}" --arg token "${entry_token}" --arg version "${kernel_version}" \
+		'.[] | select(.id | test("^" + $token + "-" + $version + "-" + $snapshot + ".conf$")) | .path' < "$entryfile" |
+	while IFS= read -r path; do
```
lnussel, line 276: you didn't test this

TobiPeterG: I did, I copied the code, set the necessary variables and echo'ed the output 

Have I missed something?

TobiPeterG: I did, I copied the code, set the necessary variables and echo'ed the output 

Have I missed something?

TobiPeterG: Ah now I know what you mean, fixed it (and tested it properly) :)

sdbootutil
```diff
@@ -240,13 +240,56 @@ sedrootflags()
...
+	fi
+
+	while IFS=$'\t' read -r id path; do
+		if [ -f "$path" ]; then
+			conf="$path"
+			conf_id="$id"
+			return 0
+		fi
+	done < <(jq -r --arg snapshot "${snapshot}" --arg token "${entry_token}" --arg version "${kernel_version}" \
+		'.[] | select(.id | test("^" + $token + "-" + $version + "-" + $snapshot + ".conf$")).id as $id | "\($id)\t\(.path)"' < "$entryfile")
```
lnussel, line 281:
```
local id="$entry_token-$kernel_version-$snapshot.conf"
...
jq -r --arg id "$id" '.[]| select(.id == $id)|.path'
```
However, you are using variables here that are not passed as argument and are not global either.

Since the function is used to find the config file in order to parse it for initrd settings, it may just as well do the parsing and return the initrds found?

sdbootutil
```diff
@@ -240,13 +240,56 @@ sedrootflags()
...
+
 remove_kernel()
 {
 	local snapshot="${1:?}"
 	local subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
 	local kernel_version="$2"
-	local id="$entry_token-$kernel_version-$snapshot.conf"
-	run_command_output bootctl unlink "$id"
+	find_conf_file "$snapshot"
+	run_command_output bootctl unlink "$conf_id"
```
lnussel, line 292: IIRC unlink doesn't need the counter. Just pass the id without counter and it would find one that has a counter

lnussel (changes requested):

sdbootutil
```diff
@@ -240,13 +240,56 @@ sedrootflags()
...
+
+update_entries_for_this_system()
+{
+	update_entries jq "[.[]|select(has(\"options\"))|select(.options|match(\"root=UUID=$root_uuid\"))]"
+}
+
+find_conf_file() {
+	local snapshot="${1:?}"
+
+	if [ ! -s "$entryfile" ]; then
```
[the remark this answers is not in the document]
TobiPeterG: It's just copied from the list_entries function. Is the check not neccessary here?

sdbootutil
TobiPeterG: oh you're right, the ID shouldn't change :)

sdbootutil
TobiPeterG: I guess the function could be used to also find initrds, but this would be two seperate tasks for one function (identify entry file + find initrds). For reusability purposes I only included finding configuration files.

However, if wanted, I can remove the function and move the logic to the reuse_initrd function as this is currently the only one using the function.

lnussel: it's fine to only limit it to the path for now for simplicity I think, given that the entry is parsed elsewhere too. The parameters need to be tuned to not rely on the calling scope. Ie could just pass the id and let the function return the path.
Something like local conf="$(get_entry_path "$entry_token-$kernel_version-$snapshot.conf")"

sdbootutil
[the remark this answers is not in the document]
lnussel: it's not useful either way I guess. It means the list wouldn't get updated when the file exists.

sdbootutil
```diff
@@ -240,6 +240,47 @@ sedrootflags()
...
+update_entries_for_this_system()
+{
+	update_entries jq "[.[]|select(has(\"options\"))|select(.options|match(\"root=UUID=$root_uuid\"))]"
+}
+
+find_conf_file() {
+	local id="${1:?}"
+
+	if [ ! -s "$entryfile" ]; then
+		update_entries_for_snapshot "$snapshot"
```
lnussel, line 271: $snapshot is not an argument of this function

sdbootutil
```diff
@@ -471,7 +508,7 @@ install_kernel()
 		done
 	fi
 	if [ -z "$failed" ]; then
-		loader_entry="$boot_root/loader/entries/$entry_token-$kernel_version-$snapshot.conf"
+		loader_entry="$boot_root/loader/entries/$entry_token-$kernel_version-$snapshot+3.conf"
```
lnussel, line 511: kernel-install reads the counter from /etc/kernel/tries, we need to mirror this

lnussel (changes requested):

sdbootutil
TobiPeterG: ah you're right. I split the parameters into kernel_version and snapshot again to be able to get the entries for the snapshots easily and construct the id with them :)

sdbootutil
TobiPeterG: You're right, done :)

sdbootutil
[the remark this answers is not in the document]
TobiPeterG: removed :)

sdbootutil
```diff
@@ -514,7 +551,20 @@ install_kernel()
 		done
 	fi
 	if [ -z "$failed" ]; then
-		loader_entry="$boot_root/loader/entries/$entry_token-$kernel_version-$snapshot.conf"
+		local tries=0
+		if [ -f /etc/kernel/tries ]; then
+			tries=$(< /etc/kernel/tries head -n 1 | tr -d '\n')
```
lnussel, line 556: use read

sdbootutil
```diff
@@ -514,7 +551,20 @@ install_kernel()
...
+		fi
+
+		if ! [[ "$tries" =~ ^[0-9]+$ ]]; then
+    		tries=0
```
lnussel, line 560: `tries=`

sdbootutil
```diff
@@ -514,7 +551,20 @@ install_kernel()
 		done
 	fi
 	if [ -z "$failed" ]; then
-		loader_entry="$boot_root/loader/entries/$entry_token-$kernel_version-$snapshot.conf"
+		local tries=0
```
lnussel, line 554: `local tries`

sdbootutil
```diff
@@ -514,7 +551,20 @@ install_kernel()
...
+		if [ -f /etc/kernel/tries ]; then
+			tries=$(< /etc/kernel/tries head -n 1 | tr -d '\n')
+		fi
+
+		if ! [[ "$tries" =~ ^[0-9]+$ ]]; then
+    		tries=0
+		fi
+
+		loader_entry="$boot_root/loader/entries/$entry_token-$kernel_version-$snapshot"
```
lnussel, line 563: `loader_entry="$boot_root/loader/entries/$entry_token-$kernel_version-$snapshot${tries:+-$tries}.conf"`

lnussel (changes requested):
needs rebase

sdbootutil
TobiPeterG: all done :)

TobiPeterG:
rebased as well :)

sdbootutil
```diff
@@ -518,7 +555,16 @@ install_kernel()
...
 	fi
 	if [ -z "$failed" ]; then
-		loader_entry="$boot_root/loader/entries/$entry_token-$kernel_version${snapshot:+-$snapshot}.conf"
+		local tries
+		if [ -f /etc/kernel/tries ]; then
+			read -r tries < /etc/kernel/tries
+		fi
+
+		if ! [[ "$tries" =~ ^[0-9]+$ ]]; then
+    		tries=
```
lnussel, line 564: missing tab? You could also do that in one line btw

sdbootutil
```diff
@@ -518,7 +555,16 @@ install_kernel()
...
+		fi
+
+		loader_entry="$boot_root/loader/entries/$entry_token-$kernel_version-$snapshot${tries:++$tries}.conf"
```
lnussel, line 567: here you are reverting the optional snapshot

sdbootutil
```diff
@@ -349,37 +390,33 @@ set_snapper_title_and_sortkey()
...
-	if [ -e "$conf" ]; then
-		local k
-		local v
-		while read -r k v; do
-			[ "$k" = 'initrd' ] || continue
-			log_info "found existing initrd $v"
-			dstinitrd+=("$v")
-		done < "$conf"
-		[ -z "$dstinitrd" ] || return 0
+	conf=$(find_conf_file "$kernel_version" "${snapshot}")
```
lnussel, line 398: should have extra quotiong around the `$(..)`. I assume shellcheck complains about that. Same below

sdbootutil
```diff
@@ -349,37 +390,33 @@ set_snapper_title_and_sortkey()
...
 
-	# check if we can reuse the initrd from the parent
-	# to avoid expensive regeneration
-	detect_parent "$subvol"
-	local parent_conf="$boot_root/loader/entries/$entry_token-$kernel_version-$parent_snapshot.conf"
-	if [ -n "$parent_subvol" ] && [ -e "$parent_conf" ]; then
-		#subvol_is_ro "$parent_subvol" || err "Parent snapshot $parent_snapshot is not read-only, can't reuse initrd"
-		local k
-		local v
+	if [ $find_conf_status -eq 0 ]; then
```
lnussel, line 411: also quoting

sdbootutil
```diff
@@ -349,37 +390,33 @@ set_snapper_title_and_sortkey()
...
+	conf=$(find_conf_file "$kernel_version" "${snapshot}")
+	local find_conf_status=$?
+
+	if [ $find_conf_status -ne 0 ]; then
+		# check if we can reuse the initrd from the parent
+		# to avoid expensive regeneration
+		detect_parent "$subvol"
+		if [ -n "$parent_subvol" ]; then
+			conf=$(find_conf_file "$kernel_version" "$parent_snapshot")
+			find_conf_status=$?
```
lnussel, line 407: quoting

lnussel (changes requested):

sdbootutil
TobiPeterG: whops, no, was just a mix of tabs and spaces, so it looked right for me. Fixed :)

sdbootutil
TobiPeterG: fixed

sdbootutil
TobiPeterG: Well, my shellcheck doesn't complain, added them

sdbootutil
TobiPeterG: done

sdbootutil
TobiPeterG: done

PR #54, opened 2024-02-29
Misc fixes

lnussel:


sdbootutil
```diff
@@ -1194,11 +1182,11 @@ pcrlock_cmdline_initrd() {
 	# 710-kernel-cmdline-initrd-entry.pcrlock.d is not part of the
 	# pcrlock standards
 	echo "$cmdline" > "$tmpdir/cmdline"
-	run_log /usr/lib/systemd/systemd-pcrlock \
+	/usr/lib/systemd/systemd-pcrlock \
```
aplanas, line 1185: should the stdout be redirected to /dev/null in all the calls? This will break the screen

sdbootutil
```diff
@@ -1226,10 +1226,19 @@ generate_tpm2_predictions_pcrlock()
 	/usr/lib/systemd/systemd-pcrlock lock-firmware-config
 	# If secure boot is disabled, this can fail.  There is patch
 	# for the policy generation, and for the authority is planned
-	/usr/lib/systemd/systemd-pcrlock lock-secureboot-policy || true
-	/usr/lib/systemd/systemd-pcrlock lock-secureboot-authority || true
-	# TODO: what drive should be here?
-	# /usr/lib/systemd/systemd-pcrlock lock-gpt "$drive"
+	/usr/lib/systemd/systemd-pcrlock lock-secureboot-policy || \
+		/usr/lib/systemd/systemd-pcrlock lock-secureboot-policy --force || true
```
aplanas, line 1230: this --force is not part of 256?

sdbootutil
```diff
@@ -1226,10 +1226,19 @@ generate_tpm2_predictions_pcrlock()
...
+		/usr/lib/systemd/systemd-pcrlock lock-secureboot-authority --force || true
+	# uses / by default
+	/usr/lib/systemd/systemd-pcrlock lock-gpt
+
+	if [ -e "$boot_root/loader/loader.conf" ]; then
+		/usr/lib/systemd/systemd-pcrlock \
+			lock-raw /boot/efi/loader/loader.conf \
+			--pcr=5 \
+			--pcrlock=/var/lib/pcrlock.d/641-sdboot-loader-conf.pcrlock
+	fi
```
aplanas, line 1241: for consistency this should be done later, at least after the shim

sdbootutil
```diff
@@ -1317,7 +1317,12 @@ generate_tpm2_predictions_pcr_oracle()
 	select_entries_for_prediction
 
 	all_pcrs=$(get_pcrs)
-	[ -z "$all_pcrs" ] || rm -f /etc/systemd/tpm2-pcr-signature.json
+	if [ -z "$all_pcrs" ]; then
+		warn "PCR Oracle configured but not enrolled in any LUKS header"
+		return 0
+	fi
+
+	/etc/systemd/tpm2-pcr-signature.json
```
aplanas, line 1325: ?

missing rm?

sdbootutil
```diff
@@ -1355,13 +1360,13 @@ generate_tpm2_predictions()
...
-	[ -e /etc/sysconfig/fde-tools ] || return 0
-
-	. /etc/sysconfig/fde-tools
 
 	if is_pcr_oracle; then
 		generate_tpm2_predictions_pcr_oracle
 	elif have_pcrlock; then
+		[ -e /etc/sysconfig/fde-tools ] || return 0
+		. /etc/sysconfig/fde-tools
+
```
aplanas, line 1369: as commented I do not like this, we should have a single PCR list as a source or truth, so it should be shared between both solutions

aplanas (changes requested):

sdbootutil
lnussel: true

sdbootutil
lnussel: k

sdbootutil
lnussel: I think the output of pcrlock needs to be reduced upstream. it's too verbose by default

sdbootutil
aplanas: lets remove it then

sdbootutil
aplanas: it is both: too verbose and too short, but the point still remains, should be > /dev/null maybe?

lnussel: I don't think so. the reason I removed the logging was that I didn't see any output at all.

aplanas: right, this is why I added it, so I can add the debug envar and see the full output of pcrlock without breaking the screen

this change will break the screen in debug mode and in normal mode

lnussel: we could set SYSTEMD_LOG_LEVEL=warning if unset for the duration of pcrlock to avoid excessive output

lnussel: What about `SYSTEMD_LOG_LEVEL=debug SYSTEMD_LOG_TARGET=journal`?

aplanas: but this does not cut the normal stdout IIUC

lnussel: That's why I thought about setting SYSTEMD_LOG_LEVEL=warning (unless SYSTEMD_LOG_LEVEL is set from outside). That way we'd see pcrlock warnings and errors on the tty while allowing to increase verbosity and log target via env vars.
In any case IMO the goal should be to reduce log output of pcrlock upstream in the first place. Those informative messages kind of make sense in the automatic mode it has right now. If we force it to  use specified pcrs however, there's not reason for pcrlock to report that it actually did what it was asked to do :)

aplanas: ok then good for me to set `SYSTEMD_LOG_LEVEL=warning`!

PR #55, opened 2024-03-02
Use new os-release and machine-id files

TobiPeterG:
This fixes an issue where the os-release file has been updated, but the old one is still used by sdbootutil to generate the new entries.
This should fix https://github.com/openSUSE/sdbootutil/issues/49

A quick test on tumbleweed worked

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -284,10 +284,45 @@ sedrootflags()
 	sed "${sed_arguments[@]}"
 }
 
+settle_entry_token()
+{
+	local snapshot="$1"
+	set_os_release "${snapshot}"
+	set_machine_id "${snapshot}"
+	[ -n "$machine_id" ] || err "Couldn't determine machine-id"
+	case "$arg_entry_token" in
+		""|auto)
+			if [ -s '/etc/kernel/entry-token' ]; then
+				read -r entry_token < '/etc/kernel/entry-token'
+			else
+				entry_token="$machine_id"
+				# bootctl has more here in case the machine id
+				# is random falls back to trying IMAGE_ID and
+				# ID, only them machine id. We assume there is
+				# a valid machine-id
+			fi
+			;;
+		machine-id) entry_token="$machine_id" ;;
+		os-id)
+			# shellcheck disable=SC2154
+			entry_token="$os_release_ID"
+			[ -n "$entry_token" ] || err "Missing ID"
+			;;
+		os-image)
+			# shellcheck disable=SC2154
+			entry_token="$os_release_IMAGE_ID"
+			[ -n "$entry_token" ] || err "Missing IMAGE_ID"
+			;;
+		*) entry_token="$arg_entry_token" ;;
+	esac
+	return 0
+}
+
 remove_kernel()
 {
 	local snapshot="$1"
 	local kernel_version="$2"
+	settle_entry_token "${snapshot}"
 	local id="$entry_token-$kernel_version${snapshot:+-$snapshot}.conf"
 	run_command_output bootctl unlink "$id"
 
@@ -346,11 +381,50 @@ set_snapper_title_and_sortkey()
 	sort_key="snapper-$sort_key"
 }
 
+set_os_release()
+{
+	local snapshot="$1"
+	local subvol=""
+	[ -z "$snapshot" ] || subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
+	os_release_files=(
+		"${subvol#"${subvol_prefix}"}/usr/lib/os-release"
+		"${subvol#"${subvol_prefix}"}/etc/os-release"
+	)
+
+	for file in "${os_release_files[@]}"; do
+		[ -f "$file" ] || continue
+		eval $(sed -ne '/^[A-Z_]\+=/s/^/os_release_/p' < "$file")
+		break
+	done
+}
+
+set_machine_id()
+{
+	local snapshot="$1"
+	local subvol=""
+	[ -z "$snapshot" ] || subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
+	machine_id_files=()
+	if is_transactional; then
+		[ -n "$snapshot" ] && machine_id_files+=("/var/lib/overlay/$snapshot/etc/machine-id")
+	fi
+	machine_id_files+=(
+		"${subvol#"${subvol_prefix}"}/etc/machine-id"
+	)
+
+	for file in "${machine_id_files[@]}"; do
+		if [ -s "$file" ]; then
+			read -r machine_id < "$file"
+			break
+		fi
+	done
+}
+
 reuse_initrd() {
 	local snapshot="$1"
 	local subvol="$2"
 
 	[ -z "$arg_no_reuse_initrd" ] || return 1
+	settle_entry_token "${snapshot}"
 
 	local conf="$boot_root/loader/entries/$entry_token-$kernel_version-$snapshot.conf"
 	if [ -e "$conf" ]; then
@@ -367,6 +441,9 @@ reuse_initrd() {
 	# check if we can reuse the initrd from the parent
 	# to avoid expensive regeneration
 	detect_parent "$subvol"
+	if [ -n "$parent_snapshot" ]; then
+		settle_entry_token "$parent_snapshot"
+	fi
 	local parent_conf="$boot_root/loader/entries/$entry_token-$kernel_version-$parent_snapshot.conf"
 	if [ -n "$parent_subvol" ] && [ -e "$parent_conf" ]; then
 		#subvol_is_ro "$parent_subvol" || err "Parent snapshot $parent_snapshot is not read-only, can't reuse initrd"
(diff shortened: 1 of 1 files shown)


sdbootutil
```diff
@@ -1427,7 +1427,38 @@ eval "$(bootctl 2>/dev/null | sed -ne 's/Firmware Arch: *\(\w\+\)/firmware_arch=
...
 root_subvol=$(btrfs subvol show / 2>/dev/null|head -1)
 subvol_prefix="${root_subvol%/.snapshots/*}"
-[ -s /etc/machine-id ] && read -r machine_id < /etc/machine-id
+
+root_snapshot="${root_subvol#${subvol_prefix}/.snapshots/}"
+root_snapshot="${root_snapshot%/snapshot}"
+
+os_release_files=(
+	"${subvol#${subvol_prefix}}/usr/lib/os-release"
+	"${subvol#${subvol_prefix}}/etc/os-release"
```
lnussel, line 1436: subvol is local to some fuctions. I don't think that works. We'd need to defer evaluating os-relase to when we actually need it I think. Looking at the code that might mean calling bootctl twice as there's information such as the ESP that is hopefully generic and others like the entry token that might be specific to a given snapshot

TobiPeterG: Whoops, you're right.
I thought I could use the root snapshot, but this doesn't make any sense at all.
Sorry.
I'll think about a new solution :)

TobiPeterG: I've extracted the os_release and machine_id read part into their own functions that are called before they are accessed. However, this is a naive implementation and requires a dev to make sure to update the os_release and machine_id before using them.

However, this should now *actually* solve the issue :)

sdbootutil
```diff
@@ -302,6 +302,45 @@ set_snapper_title_and_sortkey()
...
 
+set_os_release()
+{
+	local snapshot="${1:?}"
+	local subvol="${subvol_prefix}/.snapshots/${1:?}/snapshot"
+	os_release_files=(
+		"${subvol#${subvol_prefix}}/usr/lib/os-release"
+		"${subvol#${subvol_prefix}}/etc/os-release"
+		"/usr/lib/os-release"
+		"/etc/os-release"
```
lnussel, line 313: passing the snapshot is mandatory and a snapshot must have an os-release file so the fallback to / shouldn't be needed, right?

sdbootutil
```diff
@@ -302,6 +302,45 @@ set_snapper_title_and_sortkey()
...
+{
+	local snapshot="${1:?}"
+	local subvol="${subvol_prefix}/.snapshots/${1:?}/snapshot"
+	machine_id_files=()
+	if is_transactional; then
+		machine_id_files+=("/var/lib/overlay/$snapshot/etc/machine-id")
+	fi
+	machine_id_files+=(
+		"${subvol#${subvol_prefix}}/etc/machine-id"
+		"/etc/machine-id"
```
lnussel, line 333: same here

sdbootutil
```diff
@@ -427,6 +466,8 @@ install_kernel()
 			((++i))
 		done
 	fi
+	
+	set_os_release $snapshot
```
lnussel, line 470: quoting. Try shellcheck :-) If you are on vim, I'm using vim-plugin-airline and vim-plugin-ale (and vim-plugin-editorconfig)

lnussel: also, I think we may want to call settle_entry_token exclusively here

sdbootutil
```diff
@@ -437,6 +478,8 @@ install_kernel()
 	if ! is_transactional  && subvol_is_ro "$subvol"; then
 		set_snapper_title_and_sortkey "$snapshot"
 	fi
+	
+	set_machine_id $snapshot
```
lnussel, line 482: quoting

sdbootutil
```diff
@@ -1027,6 +1068,9 @@ install_sdboot()
 
 settle_entry_token()
 {
+	local snapshot="${1:?}"
+	set_os_release $snapshot
+	set_machine_id $snapshot
```
lnussel, line 1073: quoting.
If we call settle_entry_token as mentioned above, it could probably mandate that eg. machine_id is set before it gets called

lnussel (changes requested):

sdbootutil
TobiPeterG: In theory, you are right. I thought about cases where the user maybe made an oopsy and the file has been deleted in the snapshot or isn't there for some reason.
Also, I wanted to make it as robust as possible. But I can remove them if wanted :)

sdbootutil
TobiPeterG: I use VS Code and installed a bash plugin, but that wasn't really helpful. Installed shellcheck now, this actually displays anything useful now :)
Fixed quoting, call settle_entry_token here now instead :)

sdbootutil
TobiPeterG: I am a little unsure which call to action results from this tbh :sweat_smile:

sdbootutil
```diff
@@ -1422,26 +1464,22 @@ case "$1" in
...
+
+root_snapshot="${root_subvol#${subvol_prefix}/.snapshots/}"
+root_snapshot="${root_snapshot%/snapshot}"
 
 if [ -n "$arg_esp_path" ] && [ "$boot_root" != "$arg_esp_path" ]; then
 	err "mismatch of esp path"
 fi
 [ -n "$arg_arch" ] && firmware_arch="$arg_arch"
-settle_entry_token
+settle_entry_token "$root_snapshot"
```
lnussel, line 1482: I guess we don't need it here anymore then

sdbootutil
lnussel: without os-release the snapshot would be broken

lnussel (changes requested):

sdbootutil
TobiPeterG: A few lines below we check if entry_token exists. So currently needed, but is the check still useful?

sdbootutil
TobiPeterG: confirmed, a system without os-release file doesn't boot :D

Removed it :)

sdbootutil

sdbootutil
lnussel: `$snapshot -> "$snapshot"`
or
`[ -n "$machine_id" ] || err ..`

sdbootutil
```diff
@@ -362,6 +398,40 @@ umount_etc()
 	umount "${snapshot_dir}/etc"
 }
 
+settle_entry_token()
+{
+	local snapshot="${1:?}"
+	set_os_release "$snapshot"
+	set_machine_id "$snapshot"
+	case "$arg_entry_token" in
+		"") [ -n "$entry_token" ] || entry_token="$machine_id" ;;
```
lnussel, line 407: Not sure why I wrote it like this originally. Now it seems empty should just be `auto`

sdbootutil
TobiPeterG: the quoting is already fixed :)

TobiPeterG: added the check :)

sdbootutil
TobiPeterG: removed the check, added settle_entry_token in every function where we use the $entry_token variable

sdbootutil
TobiPeterG: well, I can't remove the line `"") [ -n "$entry_token" ] || entry_token="$machine_id" ;;` as then e.g remove-all-kernels fails, so it seems to do something

lnussel: I mean merging it with the "auto" branch
`""|auto)`

lnussel:
looks ok but needs rebase

TobiPeterG:
quick question: In line 429 and 427, -$snapshot is used, even though it might not be set. Is this intended?
e.g.: local conf="$boot_root/loader/entries/$entry_token-$kernel_version-$snapshot.conf"

lnussel:
> quick question: In line 429 and 427, -$snapshot is used, even though it might not be set. Is this intended? e.g.: local conf="$boot_root/loader/entries/$entry_token-$kernel_version-$snapshot.conf"

Should most likely be `${snapshot:--$snapshot}` always

sdbootutil
```diff
@@ -346,11 +381,50 @@ set_snapper_title_and_sortkey()
...
+		eval $(sed -ne '/^[A-Z_]\+=/s/^/os_release_/p' < "$file")
+		break
+	done
+}
+
+set_machine_id()
+{
+	local snapshot="$1"
+	local subvol=""
+	[ -n "$snapshot" ] && subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
```
lnussel, line 405: should be `assert()` style logic in one liners. Ie
`[ -z "$something" ] || ...`

lnussel (changes requested):

sdbootutil
TobiPeterG: done

lnussel:
https://github.com/openSUSE/sdbootutil/pull/78

Issue #57, opened 2024-03-04
Bug: Installing multiple kernels sometimes fails


TobiPeterG:
The installation of multiple kernels sometimes fails with
`dracut[F]: Will not override existing initramfs (/tmp/sdboot.MJfULI/initrd-0) without --force`

However, installing the kernel manually afterwards (using sdbootutil -> kernels -> install) works.

PR #58, opened 2024-03-04
Delete correct initrd file after installation

TobiPeterG:
Otherwise the file will remain orphaned in the tmpdir and may cause dracut to fail on subsequent builds

closes https://github.com/openSUSE/sdbootutil/issues/57

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -463,6 +463,7 @@ install_kernel()
 		while [ -e "$tmpdir/initrd-$i" ]; do
 			if [ ! -e "$boot_root${dstinitrd[$i]}" ]; then
 				install_with_rollback "$tmpdir/initrd-$i" "$boot_root${dstinitrd[$i]}" || { failed=initrd; break; }
+				rm -f "$tmpdir/initrd-$i"
 			fi
 			((++i))
 		done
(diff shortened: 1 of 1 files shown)


TobiPeterG:
This was at least the change that fixed my builds. Please check if the behaviour is correct now :)

lnussel:
we could just add the rm to install_with_rollback I think

TobiPeterG:
> we could just add the rm to install_with_rollback I think

Above, install_with_rollback is called with the kernel image so here the source shouldn't be deleted.
I could add the rm commands directly after the install_with_rollback calls or think of a different solution (e.g. adding a function install_with_rollback_and_delete that calls install_with_rollback internally and remove the source file afterwards)

lnussel:
right. let's put the rm in the existing loop then

TobiPeterG:
> right. let's put the rm in the existing loop then

done :)
They only get removed though if $failed is not set. Is this an issue?

lnussel:
> > right. let's put the rm in the existing loop then
> 
> done :) They only get removed though if $failed is not set. Is this an issue?

when adding multiple kernels that would mean the second could pick up files from the first if the first one fails. We should move all files to a subdir of $tmpdir and delete that one afterwards I guess

Issue #35, opened 2023-12-18
mkinitrd command


lnussel:
` sdbootutil --no-reuse-initrd add-all-kernels` is annyoing to type. add `sdbootutil mkinitrd` for convenience.

PR #59, opened 2024-03-04
Add add-all-kernels-clean function

TobiPeterG:
This allows to easily add all kernels and regenerate the initrds for them
Fixes https://github.com/openSUSE/sdbootutil/issues/35

The name of the function is still quite long, but is easier to understand.
If a shorter name is preferred, this can be changed ofc :)

Command tested on Tumbleweed


lnussel:
just mkinitrd? even if it also installs kernels?

TobiPeterG:
> just mkinitrd? even if it also installs kernels?

renamed it :)

Issue #61, opened 2024-03-05
Enhancement: Add bash completions


TobiPeterG:
Whenever I enter a command using sdbootutil, I have to know the exact command and don't get any hints which commands exist. This makes entering commands difficult. Other tools have recommendations for available commands.
It would be nice if sdbootutil offered this as well.

PR #62, opened 2024-03-05
Add bash completions

TobiPeterG:
This will complete started commands and provide a list of possible commands and options.

Closes https://github.com/openSUSE/sdbootutil/issues/61

Unfortunately, we can't access the available snapshots, so the user still has to know which ones are available on his system. Equally, we can't enter snapshots to find kernel versions, so we recommend only the currently installed kernels.

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -2854,9 +2854,8 @@ if [ -z "$SYSTEMD_LOG_LEVEL" ] && [ -n "$verbose" ]; then
 fi
 
 define_commands
-if [ -z "$1" ] || [ "${commands["$1"]}" = "interactive" ]; then
-	stty_size
-	interactive=1
+if [ -z "$1" ]; then
+	helpandquit
 elif [ -z ${commands["$1"]+yes} ]; then
 	err "unknown command $1"
 fi
--- a/sdbootutil
+++ b/sdbootutil
@@ -2805,6 +2805,14 @@ bootloader_name()
 
 ####### main #######
 
+if [ "$1" = "_print_bash_completion_data" ]; then
+	declare -f set_image_name
+	declare -f eval_bootctl
+	declare -f define_commands
+	declare -f define_options
+	exit 0
+fi
+
 define_options
 getopt_tmp=$(getopt -o hc:v --long "$opts_long" -n "${0##*/}" -- "$@")
 eval set -- "$getopt_tmp"
--- a/sdbootutil
+++ b/sdbootutil
@@ -2718,6 +2718,12 @@ set_image_name() {
 	fi
 }
 
+eval_bootctl()
+{
+	# XXX: bootctl should have json output for that too
+	eval "$(bootctl 2>/dev/null | sed -ne 's/Firmware Arch: *\(\w\+\)/firmware_arch="\1"/p;s/ *token: *\(\w\+\)/entry_token="\1"/p;s, *\$BOOT: *\([^ ]\+\).*,boot_root="\1",p')"
+}
+
 bootloader_name()
 {
 	if is_sdboot "${1-$root_snapshot}"; then
@@ -2777,9 +2783,7 @@ esac
 
 [ -n "$arg_esp_path" ] && export SYSTEMD_ESP_PATH="$arg_esp_path"
 
-# XXX: bootctl should have json output for that too
-# shellcheck disable=SC2016
-eval "$(bootctl 2>/dev/null | sed -ne 's/Firmware Arch: *\(\w\+\)/firmware_arch="\1"/p;s/ *token: *\(\w\+\)/entry_token="\1"/p;s, *\$BOOT: *\([^ ]\+\).*,boot_root="\1",p')"
+eval_bootctl
 read -r root_uuid root_device < <(findmnt / -v -r -n -o UUID,SOURCE)
 root_device_is_crypt=
 [ "$(lsblk --noheadings -o TYPE "$root_device")" = "crypt" ] && root_device_is_crypt=1
--- a/sdbootutil
+++ b/sdbootutil
@@ -2703,6 +2703,21 @@ unenroll()
 	done
 }
 
+set_image_name() {
+	[ -z "$image" ] || return
+
+	declare -gA arch_image_map=(
+		[x64]="vmlinuz"
+		[aa64]="Image"
+	)
+
+	if [ -n "${arch_image_map[$firmware_arch]}" ]; then
+		image="${arch_image_map[$firmware_arch]}"
+	else
+		err "Unsupported architecture $firmware_arch"
+	fi
+}
+
 bootloader_name()
 {
 	if is_sdboot "${1-$root_snapshot}"; then
@@ -2796,11 +2811,7 @@ fi
 [ -n "$root_subvol" ] || [ -z "$have_snapshots" ] || err "Can't determine root subvolume"
 [ -n "$root_device" ] || err "Can't determine root device"
 [ -n "$firmware_arch" ] || err "Can't determine firmware arch"
-case "$firmware_arch" in
-	x64) image=vmlinuz ;;
-	aa64) image=Image ;;
-	*) err "Unsupported architecture $firmware_arch" ;;
-esac
+set_image_name
 
 # shellcheck disable=SC1091
 [ -e /etc/sysconfig/bootloader ] && . /etc/sysconfig/bootloader
--- a/sdbootutil
+++ b/sdbootutil
@@ -2756,6 +2756,42 @@ define_commands() {
 	)
 }
 
+define_options() {
+	declare -gA options_with_arg=(
+		[help]=""
+		[verbose]=""
+		[esp-path]="_path"
+		[entry-token]="_path"
+		[arch]="_arch_name"
+		[image]="_image_name"
+		[entry-keys]="_find_kernels"
+		[no-variables]=""
+		[no-reuse-initrd]=""
+		[no-random-seed]=""
+		[all]=""
+		[sync]=""
+		[portable]=""
+		[removable]=""
+		[only-default]=""
+		[default-snapshot]=""
+		[ask-key]=""
+		[ask-pin]=""
+		[ask-pw]=""
+		[method]="_method"
+		[signed-policy]=""
+		[pcr]="_none"
+	)
+	opts_long=""
+	for opt in "${!options_with_arg[@]}"; do
+		if [ "${options_with_arg[$opt]}" ]; then
+			opts_long+="${opt}:,"
+		else
+			opts_long+="${opt},"
+		fi
+	done
+	opts_long="${opts_long%,}"
+}
+
 bootloader_name()
 {
 	if is_sdboot "${1-$root_snapshot}"; then
@@ -2769,8 +2805,9 @@ bootloader_name()
 
 ####### main #######
 
-getopttmp=$(getopt -o hc:v --long help,verbose,esp-path:,entry-token:,arch:,image:,entry-keys:,no-variables,no-reuse-initrd,no-random-seed,all,sync,portable,removable,only-default,default-snapshot,ask-key,ask-pin,ask-pw,method:,signed-policy,pcr: -n "${0##*/}" -- "$@")
-eval set -- "$getopttmp"
+define_options
+getopt_tmp=$(getopt -o hc:v --long "$opts_long" -n "${0##*/}" -- "$@")
+eval set -- "$getopt_tmp"
 
 while true ; do
 	case "$1" in
--- a/sdbootutil
+++ b/sdbootutil
@@ -2724,6 +2724,38 @@ eval_bootctl()
 	eval "$(bootctl 2>/dev/null | sed -ne 's/Firmware Arch: *\(\w\+\)/firmware_arch="\1"/p;s/ *token: *\(\w\+\)/entry_token="\1"/p;s, *\$BOOT: *\([^ ]\+\).*,boot_root="\1",p')"
 }
 
+define_commands() {
+	declare -gA commands=(
+		[install]=""
+		[needs-update]=""
+		[update]=""
+		[force-update]=""
+		[add-kernel]="kernel"
+		[remove-kernel]="kernel"
+		[set-default-snapshot]=""
+		[add-all-kernels]=""
+		[mkinitrd]=""
+		[remove-all-kernels]=""
+		[is-installed]=""
+		[list-snapshots]=""
+		[list-entries]=""
+		[list-kernels]=""
+		[list-devices]=""
+		[show-entry]="kernel"
+		[update-entry]="kernel"
+		[update-all-entries]=""
+		[is-bootable]=""
+		[set-default]="id"
+		[get-default]=""
+		[set-timeout]="seconds"
+		[get-timeout]=""
+		[enroll]=""
+		[unenroll]=""
+		[update-predictions]=""
+		[bootloader]=""
+	)
+}
+
 bootloader_name()
 {
 	if is_sdboot "${1-$root_snapshot}"; then
@@ -2776,10 +2808,13 @@ if [ -z "$SYSTEMD_LOG_LEVEL" ] && [ -n "$verbose" ]; then
 	export SYSTEMD_LOG_LEVEL
 fi
 
-case "$1" in
-	install|needs-update|update|force-update|add-kernel|remove-kernel|set-default-snapshot|add-all-kernels|mkinitrd|remove-all-kernels|is-installed|list-snapshots|list-entries|list-kernels|list-devices|show-entry|update-entry|update-all-entries|is-bootable|set-default|get-default|set-timeout|get-timeout|enroll|unenroll|update-predictions|bootloader|"") ;;
-	*) err "unknown command $1" ;;
-esac
+define_commands
+if [ -z "$1" ] || [ "${commands["$1"]}" = "interactive" ]; then
+	stty_size
+	interactive=1
+elif [ -z ${commands["$1"]+yes} ]; then
+	err "unknown command $1"
+fi
 
 [ -n "$arg_esp_path" ] && export SYSTEMD_ESP_PATH="$arg_esp_path"
 
--- a/sdbootutil.spec
+++ b/sdbootutil.spec
@@ -94,6 +94,17 @@ JEOS module for full disk encryption enrollment. The module
 present the different options and delegate into sdbootutil-enroll
 service the effective enrollment.
 
+%package bash-completion
+Summary:        Bash completions for sdbootutil
+Requires:       sdbootutil >= %{version}-%{release}
+Requires:       bash
+Requires:       bash-completion
+
+%description bash-completion
+Bash completions script for sdbootutil.
+Allows the user to press TAB to see available commands,
+options and parameters.
+
 %prep
 %setup -q
 
@@ -123,6 +134,9 @@ install -D -m 755 10-%{name}.tukit %{buildroot}%{_prefix}/lib/tukit/plugins/10-%
 
 # kernel-install
 install -D -m 755 50-%{name}.install %{buildroot}%{_prefix}/lib/kernel/install.d/50-%{name}.install
+
+#bash completions
+install -D -m 755 completions/bash_sdbootutil %{buildroot}%{_datadir}/bash-completion/completions/sdbootutil
 
 # tmpfiles
 install -D -m 755 kernel-install-%{name}.conf \
@@ -194,5 +208,10 @@ sdbootutil update
 %{_datadir}/jeos-firstboot/modules/enroll
 %dir %{_unitdir}/jeos-firstboot.service.d
 %{_unitdir}/jeos-firstboot.service.d/jeos-firstboot-enroll-override.conf
+
+%files bash-completion
+%dir %{_datadir}/bash-completion
+%dir %{_datadir}/bash-completion/completions
+%{_datadir}/bash-completion/completions/sdbootutil
 
 %changelog
--- a/completions/bash_sdbootutil
+++ b/completions/bash_sdbootutil
@@ -1,0 +1,114 @@
+# shellcheck disable=SC2148
+shopt -s nullglob
+# shellcheck disable=SC2148
+err() {
+	:
+}
+
+_none() {
+	declare -ga result
+	result=()
+}
+
+_find_kernels()
+{
+	local fn kv
+	declare -ga result
+	result=()
+
+	for fn in /usr/lib/modules/*/"$image"; do
+		kv="${fn%/*}"
+		kv="${kv##*/}"
+		result+=("$kv")
+	done
+}
+
+_find_entry_ids()
+{
+	declare -ga result
+	result=()
+
+	mapfile -t result < <(bootctl list --json=short | jq -r '.[].id')
+}
+
+_arch_name() {
+	declare -ga result
+	result=("${!arch_image_map[@]}")
+}
+
+_image_name() {
+	declare -ga result
+	result=()
+	result=("${!arch_image_map[@]}")
+}
+
+_method() {
+	declare -ga result
+	result=()
+    for i in "tpm2" "tpm2+pin" "fido2" "password" "recovery-key"; do
+	    result+=("$i")
+    done
+}
+
+_sdbootutil_completion() {
+	eval "$(sdbootutil _print_bash_completion_data)"
+	local cur prev words cword image
+	COMPREPLY=()
+	_get_comp_words_by_ref -n : cur prev words cword
+
+	local command_found=0
+	eval_bootctl
+	set_image_name
+	define_commands
+
+	define_options
+	# shellcheck disable=SC2154
+	IFS=, read -r -a opts <<< "$opts_long"
+	local opts_with_dashes=()
+	for opt in "${opts[@]}"; do
+		opts_with_dashes+=("--${opt//:/}")
+	done
+
+	local i=0
+
+	for word in "${words[@]:1:$cword-1}"; do
+		if [ -n "${commands["$word"]+yes}" ]; then
+			command_found=1
+		fi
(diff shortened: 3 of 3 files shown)


lnussel:
I'd probably add a hidden function to sdbootutil that can be eval'd from the completion script to avoid duplicating all that code

TobiPeterG:
> I'd probably add a hidden function to sdbootutil that can be eval'd from the completion script to avoid duplicating all that code

I've looked a bit into how bash completion works and it seems like the bash completion has to happen in the same shell context. So I can A) source the sdbootutil script, but that would execute the functions that might fail as the bash completion can be called as non-super user
B) call a function in sdbootutil where I echo the recommendations, but that would also not be a nice solution
C) Create two functions in sdbootutil that are callable that return the architecture/image name and kernels in the current snapshot.

Or do you have a different idea?
I agree that the code duplication isn't nice

lnussel:
something like `eval $(sdbootutil _print_bash_completion_data)`

TobiPeterG:
> something like `eval $(sdbootutil _print_bash_completion_data)`

I've tried to implement this. I couldn't test it yet though, but please tell me if this is what you had in mind :)
The code also isn't nice yet, it's just thought as a concept currently :)

sdbootutil
```diff
@@ -1377,6 +1392,56 @@ generate_tpm2_predictions()
 	fi
 }
 
+_sdbootutil_completion()
```
lnussel, line 1395: this function still belongs into the separate completion script

sdbootutil
```diff
@@ -777,11 +777,17 @@ calc_chksum() {
 declare -A found_kernels
 find_kernels()
 {
-	local subvol="${subvol_prefix}/.snapshots/${1:?}/snapshot"
+	local prefix
 	local fn kv
 	found_kernels=()
 
-	for fn in "${subvol#${subvol_prefix}}"/usr/lib/modules/*/"$image"; do
+	if [[ -z "$1" ]]; then
```
lnussel, line 784: A single [ is sufficient here

sdbootutil
```diff
@@ -790,6 +796,15 @@ find_kernels()
 	done
 }
 
+find_image()
```
lnussel, line 799: not really finding anything. determining the kernel image name.

sdbootutil
```diff
@@ -1377,6 +1392,56 @@ generate_tpm2_predictions()
...
 
+_sdbootutil_completion() 
+{
+	local cur prev words cword subvol_prefix image
+	COMPREPLY=()
+	_get_comp_words_by_ref -n : cur prev words cword
+
+	local opts_commands="install needs-update update force-update add-kernel remove-kernel set-default-snapshot add-all-kernels remove-all-kernels is-installed list-snapshots list-entries list-kernels is-bootable update-predictions kernels snapshots entries"
+	local opts_options="--help --verbose --flicker --esp-path --arch --entry-token --image --no-variables --no-reuse-initrd --no-random-seed --all"
+	local opts_arguments="--esp-path --arch --entry-token --image"
```
lnussel, line 1403: those bits could be extracted from existing information to avoid duplicating it. eg we could put the long options of getopt into a separate variable and parse that for completion purposes.

Similarly the supported command list could be put into an associative array. The existing case statement is no longer cool I think.

lnussel (changes requested):

sdbootutil
TobiPeterG: fixed

sdbootutil

sdbootutil
TobiPeterG: renamed :)

sdbootutil
TobiPeterG: done, did a bit of testing, seems to work :)
Please check if it works for you :)

sdbootutil
```diff
@@ -776,24 +776,48 @@ calc_chksum() {
...
 {
-	local subvol="${subvol_prefix}/.snapshots/${1:?}/snapshot"
+	local prefix
 	local fn kv
+	# map with kernel version as key and checksum as value
+	declare -gA found_kernels
 	found_kernels=()
 
-	for fn in "${subvol#${subvol_prefix}}"/usr/lib/modules/*/"$image"; do
+	if [ "$1" == "completion" ]; then
```
lnussel, line 787: only the $1 != completion needed here

sdbootutil
```diff
@@ -776,24 +776,48 @@ calc_chksum() {
...
+
+	for fn in "${prefix}"/usr/lib/modules/*/"$image"; do
 		kv="${fn%/*}"
 		kv="${kv##*/}"
-		calc_chksum "$fn"
+		if [ ! "$1" == "completion" ]; then
+			calc_chksum "$fn"
+			log_info "found kernel $kv = $chksum"
+		else
+			chksum="$kv"
```
lnussel, line 800: that's nasty, don't do that :) As you are only interested in the kernel versions, not the checkums. maybe it makes sense to run the loop separately in the completion script after all

sdbootutil
```diff
@@ -1397,8 +1421,27 @@ main_menu()
...
 }
 
+define_commands()
+{
+	opts_commands="install needs-update update force-update add-kernel remove-kernel set-default-snapshot add-all-kernels mkinitrd remove-all-kernels is-installed list-snapshots list-entries list-kernels is-bootable update-predictions kernels snapshots entries"
+	opts_options="--help --verbose --flicker --esp-path --arch --entry-token --image --no-variables --no-reuse-initrd --no-random-seed --all"
+	opts_arguments="--esp-path --arch --entry-token --image"
+	commands_needing_snaps="add-kernel remove-kernel set-default-snapshot add-all-kernels remove-all-kernels is-bootable list-entries list-kernels mkinitrd"
+	commands_needing_kernels="add-kernel remove-kernel"
+	commands_interactive="kernels snapshots entries"
```
lnussel, line 1431: use associative arrays!

sdbootutil
```diff
@@ -1420,11 +1463,13 @@ while true ; do
         esac
 done
 
-case "$1" in
-	install|needs-update|update|force-update|add-kernel|remove-kernel|set-default-snapshot|add-all-kernels|mkinitrd|remove-all-kernels|is-installed|list-snapshots|list-entries|list-kernels|is-bootable|update-predictions) ;;
-	kernels|snapshots|entries|"") stty_size; interactive=1 ;;
-	*) err "unknown command $1" ;;
-esac
+define_commands
+if [[ -z "$1" || " $commands_interactive " =~ " $1 " ]]; then
```
lnussel, line 1467: no way. use associative arrays

lnussel (changes requested):

sdbootutil
TobiPeterG: done :)

sdbootutil
TobiPeterG: done

sdbootutil
TobiPeterG: done
Also removed the if else at the end and replaced it with a case statement

sdbootutil
TobiPeterG: done, moved a simplified version to the completion script

completion/sdbootutil
```diff
@@ -0,0 +1,70 @@
+err() {
+	:
+}
+
+find_kernels()
```
lnussel, line 5: duplicated from sdbootutil?

sdbootutil
```diff
@@ -1498,8 +1513,33 @@ main_menu()
 	done
 }
 
+define_commands() {
+	declare -gA command_types=(
```
lnussel, line 1517:
```
  declare -A commands=(
    [install]=
    [add-kernel]=kernel
    [entries]=interactive
    ...
)
```
you can test for existance of a key eg using `[ ${commands[$cmd]+yes} ]`

sdbootutil
```diff
@@ -1521,11 +1561,13 @@ while true ; do
         esac
 done
 
-case "$1" in
-	install|needs-update|update|force-update|add-kernel|remove-kernel|set-default-snapshot|add-all-kernels|mkinitrd|remove-all-kernels|is-installed|list-snapshots|list-entries|list-kernels|is-bootable|update-predictions|bootloader) ;;
-	kernels|snapshots|entries|"") stty_size; interactive=1 ;;
-	*) err "unknown command $1" ;;
-esac
+define_commands
+if [[ -z "$1" || ${command_types["$1"]} == "command_interactive" ]]; then
```
lnussel, line 1565: `[ -z "$1" ] || [ "${commands[$1]}" = 'interactive' ]`

sdbootutil
```diff
@@ -1498,8 +1513,33 @@ main_menu()
...
+		[remove-all-kernels]="command" [is-installed]="command"
+		[list-snapshots]="command" [list-entries]="command"
+		[list-kernels]="command" [is-bootable]="command"
+		[update-predictions]="command" [kernels]="command_interactive"
+		[snapshots]="command_interactive" [entries]="command_interactive" [mkinitrd]="command"
+		[bootloader]="command"
+		[--help]="option" [--verbose]="option" [--flicker]="option"
+		[--esp-path]="argument" [--arch]="argument" [--entry-token]="argument"
+		[--image]="argument" [--no-variables]="option" [--no-reuse-initrd]="option"
+		[--no-random-seed]="option" [--all]="option"
```
lnussel, line 1530: this is duplicating the getopt argument still
super fancy code to parse it:
```
opts_long="help,flicker,verbose,esp-path:,entry-token:,arch:,image:,no-variables,no-reuse-initrd,no-random-seed,all"
declare -A opts_long_dict
parse_opts()
{
        IFS=, eval set -- \$opts_long
        eval "${@//*/opts_long_dict[&]=}"
}
parse_opts
```

completion/sdbootutil
```diff
@@ -0,0 +1,70 @@
...
+	found_kernels=()
+
+	for fn in ""/usr/lib/modules/*/"$image"; do
+		kv="${fn%/*}"
+		kv="${kv##*/}"
+		found_kernels+=("$kv")
+	done
+}
+
+eval "$(sdbootutil _print_bash_completion_data)"
```
lnussel, line 17: this should be done once inside the _sdbootutil_completion function I guess to avoid the code from running right when a new shell is started

completion/sdbootutil
```diff
@@ -0,0 +1,70 @@
...
+	eval_bootctl
+	set_image_name
+	define_commands
+	local i=0
+
+	for word in "${words[@]:1:$cword-1}"; do
+		if [[ ${command_types["$word"]} =~ ^(command|command_needing_kernel)$ ]]; then
+			command_found=1
+		fi
+		if [[ " --arch " == " $word " ]]; then
```
lnussel, line 34: no need for [[ here and many other places

lnussel (changes requested):

completion/sdbootutil
TobiPeterG: Above you said it might make sense to run the loop in the script. As the function in sdbootutil calculates the checksum etc., it wouldn't really be nice code.

sdbootutil
TobiPeterG: I don't think this is really required. The completion script just needs the option names, not the parsing,

completion/sdbootutil
TobiPeterG: done

completion/sdbootutil
TobiPeterG: done :)

sdbootutil
```diff
@@ -1610,10 +1628,49 @@ main_menu()
...
+		[snapshots]=interactive
+		[entries]=interactive
+		[mkinitrd]=""
+		[bootloader]=""
+	)
+}
+
+define_options()
+{
+	declare -g opts_long="help,flicker,verbose,esp-path,entry-token,arch,image,no-variables,no-reuse-initrd,no-random-seed,all"
```
lnussel, line 1658: lacks colons for arguments. A fancy way would be to use an associative array here too. The value could be either empty (no argument) or a function name for the completion of the argument. opts_long then consists of the keys of the array with colons appended to the keys that have a completion function defined.

lnussel (dismissed):
almost there

sdbootutil
TobiPeterG: Sorry for the delay, please check if this is now what you are looking for :)

TobiPeterG:
@aplanas can you have a look at this one please? :)

completion/sdbootutil
```diff
@@ -0,0 +1,91 @@
...
+	:
+}
+
+_find_kernels()
+{
+	local fn kv
+	declare -ga result
+	result=()
+
+	for fn in ""/usr/lib/modules/*/"$image"; do
```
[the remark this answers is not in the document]
TobiPeterG: I removed the first two quote signs, but this seems to be correct. At least shellcheck doesn't like it otherwise.

sdbootutil.spec
```diff
@@ -90,6 +90,10 @@ package may disable other plugin scripts that are incompatible.
 %install
 install -D -m 755 sdbootutil %{buildroot}%{_bindir}/sdbootutil
 
+#bash completions
+mkdir -p %{buildroot}%{_datadir}/bash-completion/completions
```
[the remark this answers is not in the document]
TobiPeterG: removed

sdbootutil
```diff
@@ -1751,57 +1834,56 @@ else
...
-elif [ "$1" = "snapshots" ]; then
-	show_snapper
-elif [ "$1" = "entries" ]; then
-	show_entries
-else
-	main_menu
-fi
+case "$1" in
+	"install")
+		install_bootloader "${2:-$root_snapshot}" ;;
```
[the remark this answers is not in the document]
TobiPeterG: done

sdbootutil
```diff
@@ -999,6 +999,29 @@ find_kernels()
 	done
 }
 
+set_image_name() {
+	if [ -n "$image" ]; then
+		return
+	fi
```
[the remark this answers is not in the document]
TobiPeterG: done

sdbootutil
```diff
@@ -999,6 +999,29 @@ find_kernels()
...
+
+	declare -gA arch_image_map=(
+		[x64]="vmlinuz"
+		[aa64]="Image"
+	)
+
+	if [[ -n "${arch_image_map[$firmware_arch]}" ]]; then
```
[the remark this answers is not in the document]
TobiPeterG: done

completion/sdbootutil
```diff
@@ -0,0 +1,91 @@
...
+		COMPREPLY=( $(compgen -W "${result[*]}" -- "$cur") )
+	elif [[ $cur == -* ]] || [ $command_found -eq 1 ]; then
+		# shellcheck disable=SC2207
+		COMPREPLY=( $(compgen -W "${opts_with_dashes[*]}" -- "$cur") )
+	elif [ $command_found -eq 0 ]; then
+		# shellcheck disable=SC2207
+		COMPREPLY=( $(compgen -W "${!commands[*]}" -- "$cur") )
+	fi
+}
+complete -F _sdbootutil_completion sdbootutil
```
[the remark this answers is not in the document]
TobiPeterG: done

sdbootutil
```diff
@@ -1669,10 +1692,74 @@ main_menu()
...
+
+	opts_long="${opts_long%,}"
+}
+
 ####### main #######
 
-getopttmp=$(getopt -o hc:v --long help,flicker,verbose,esp-path:,entry-token:,arch:,image:,entry-keys:,no-variables,no-reuse-initrd,no-random-seed,all -n "${0##*/}" -- "$@")
-eval set -- "$getopttmp"
+if [ "$1" == "_print_bash_completion_data" ]; then
+	declare -f set_image_name
```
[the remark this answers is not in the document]
TobiPeterG: Why exactly is this neccessary?
Previously, these global variables were already in the sdbootutil script set, these functions just make sure they are also accessible in the bash completion script.
We can't just export the $image variable as the options given to sdbootutil might change it or the architecture, which needs to then reset the image to the new name.
We definitely need the function available in the completion script.

completion/sdbootutil
```diff
@@ -0,0 +1,91 @@
...
+_image_name() {
+	declare -ga result
+	result=()
+	for key in "${!arch_image_map[@]}"; do
+		result+=("${arch_image_map[$key]}")
+	done
+}
+
+_sdbootutil_completion() {
+	eval "$(sdbootutil _print_bash_completion_data)"
```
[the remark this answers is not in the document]
TobiPeterG: At first I did that, Ludwig didn't like it :D

When I remember correctly, these completions scripts get executed once when a new shell is created. This would mean that sdbootutil gets called automatically by bash, even when there is no intend to call it manually as user.
This fixes that.

completion/sdbootutil
```diff
@@ -0,0 +1,91 @@
...
+	local cur prev words cword image
+	COMPREPLY=()
+	_get_comp_words_by_ref -n : cur prev words cword
```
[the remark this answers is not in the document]
TobiPeterG: Hmm. The completion script wouldn't be called without the package afaik, so not including it as required doesn't break anything I guess.

sdbootutil.spec
```diff
@@ -141,6 +145,9 @@ sdbootutil update
 %files
 %license LICENSE
 %{_bindir}/sdbootutil
+%dir %{_datadir}/bash-completion
```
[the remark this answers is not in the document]
TobiPeterG: see above.
I guess that including it in the regular package shouldn't break anything, even when bash-completion is not installed.
If you want to, though, I can of course place this in its own package. It would become a long list of packages if we support other shells in the future as well :D

completion/sdbootutil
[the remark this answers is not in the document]
aplanas: Interesting. But ""/... I parsed as "append the empty string to /..". It is not 'wrong', so shellcheck cannot complain, but I guess that it is not what you want to do?

completion/sdbootutil
```diff
@@ -0,0 +1,91 @@
...
+		result+=("$kv")
+	done
+}
+
+_arch_name() {
+	declare -ga result
+	result=()
+	for n in "${!arch_image_map[@]}"; do
+		result+=("$n")
+	done
```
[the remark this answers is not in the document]
TobiPeterG: Yes, you are right :)
Done

completion/sdbootutil
```diff
@@ -0,0 +1,91 @@
...
+}
+
+_image_name() {
+	declare -ga result
+	result=()
+	for key in "${!arch_image_map[@]}"; do
+		result+=("${arch_image_map[$key]}")
+	done
```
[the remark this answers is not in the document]
TobiPeterG: done

completion/sdbootutil
```diff
@@ -0,0 +1,91 @@
...
+	IFS=, read -r -a opts <<< "$opts_long"
+	local opts_with_dashes=()
+	for opt in "${opts[@]}"; do
+		opts_with_dashes+=("--${opt//:/}")
+	done
+
+	local i=0
+
+	for word in "${words[@]:1:$cword-1}"; do
+		if [ -n "${commands["$word"]+yes}" ]; then
```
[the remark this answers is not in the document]
TobiPeterG: I think yes.
It should match all commands with a key, not only those with a value

sdbootutil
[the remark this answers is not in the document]
aplanas: I do not follow ... the architecture cannot change on the running system. Once the function in sdbootutil knows the value of those variables, that will not change for this completion session.

completion/sdbootutil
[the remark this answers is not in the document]
TobiPeterG: No, it's not. But as it seems, we can't just use single quotes around everything as this would run only once (at least shellcheck says that)

sdbootutil.spec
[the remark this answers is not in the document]
aplanas: I think that should be a subpackage, in case that others use zsh, fish or others

completion/sdbootutil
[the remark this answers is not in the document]
aplanas: That is OK, but I was thinking in the reverse case. If this is a subpackage and a user install this subpackage, then they also want to install the bash-completion package too.

completion/sdbootutil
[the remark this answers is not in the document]
aplanas: Ah, so there is a complain for `"/usr/lib/modules/*/$image";` from shellcheck! What is the ID?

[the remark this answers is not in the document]
TobiPeterG: yes.
https://www.shellcheck.net/wiki/SC2066

sdbootutil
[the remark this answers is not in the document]
TobiPeterG: There is an option --arch to set the architecture manually and --image to set the image name manually..
This option might be used, so we need to change the image name.

sdbootutil.spec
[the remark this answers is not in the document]
TobiPeterG: done

completion/sdbootutil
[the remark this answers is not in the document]
TobiPeterG: done

aplanas:
@TobiPeterG the PR is looking good. Give me some time to test it a bit

TobiPeterG:
Should I just rebase or do you have any issues with the functionality/code that I should fix with a rebase?

aplanas:
Yes please, do a rebase. But I need to research a bit more. I found a good reference

sdbootutil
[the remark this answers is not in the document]
aplanas: ok

sdbootutil
```diff
@@ -2014,10 +2035,76 @@ main_menu()
...
 
+define_commands() {
+	declare -gA commands=(
+		[install]=""
+		[needs-update]=""
+		[update]=""
+		[force-update]=""
+		[add-kernel]=kernel
+		[remove-kernel]=kernel
+		[set-default-snapshot]=""
```
aplanas, line 2046: I think this has a parameter

sdbootutil
```diff
@@ -2014,10 +2035,76 @@ main_menu()
...
+		[add-all-kernels]=""
+		[remove-all-kernels]=""
+		[is-installed]=""
+		[list-snapshots]=""
+		[list-entries]=""
+		[list-kernels]=""
+		[show-entry]=""
```
aplanas, line 2053: I think this also has parameters

aplanas (commented):
I think we can split this PR is two: one for all the bash completions, but another one for the commands refactoring.  This last part is the main source of conflicts right now

TobiPeterG:
Now depends on #172 

I have refactored this PR and split it into multiple commits. Please have a look :)

sdbootutil
TobiPeterG: set-default-snapshot's SNAPSHOT parameter is optional

TobiPeterG:
To also support snapshots as completion, it would be the easiest to add a new option "--snapshot" that I can check for i nthe completion script. Otherwise it's impossible to know if the user wants to add an option or a snapshot number and I don't want to show both in one list.

TobiPeterG:
Rebased :)

I added the "_none" function when no completion is useful.

TobiPeterG:
@aplanas Rebased :)

aplanas (changes requested):

aplanas (changes requested):

sdbootutil
```diff
@@ -2755,16 +2853,17 @@ if [ -z "$SYSTEMD_LOG_LEVEL" ] && [ -n "$verbose" ]; then
...
 fi
 
-case "$1" in
-	install|needs-update|update|force-update|add-kernel|remove-kernel|set-default-snapshot|add-all-kernels|mkinitrd|remove-all-kernels|is-installed|list-snapshots|list-entries|list-kernels|list-devices|show-entry|update-entry|update-all-entries|is-bootable|set-default|get-default|set-timeout|get-timeout|enroll|unenroll|update-predictions|bootloader|"") ;;
-	*) err "unknown command $1" ;;
-esac
+define_commands
+if [ -z "$1" ] || [ "${commands["$1"]}" = "interactive" ]; then
+	stty_size
+	interactive=1
```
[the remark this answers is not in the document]
TobiPeterG: Yes, thanks, will do :)
Missed it, sorry.

sdbootutil
```diff
@@ -2755,16 +2853,17 @@ if [ -z "$SYSTEMD_LOG_LEVEL" ] && [ -n "$verbose" ]; then
 	export SYSTEMD_LOG_LEVEL
 fi
 
-case "$1" in
-	install|needs-update|update|force-update|add-kernel|remove-kernel|set-default-snapshot|add-all-kernels|mkinitrd|remove-all-kernels|is-installed|list-snapshots|list-entries|list-kernels|list-devices|show-entry|update-entry|update-all-entries|is-bootable|set-default|get-default|set-timeout|get-timeout|enroll|unenroll|update-predictions|bootloader|"") ;;
-	*) err "unknown command $1" ;;
-esac
+define_commands
+if [ -z "$1" ] || [ "${commands["$1"]}" = "interactive" ]; then
+	stty_size
```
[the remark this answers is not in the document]
TobiPeterG: done

sdbootutil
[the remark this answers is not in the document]
TobiPeterG: done

sdbootutil
TobiPeterG: Yes, they have the snapshot number as parameter, however, it's optional. So we shouldn't "force" the user to set it.

TobiPeterG: Ah waut, show entry, yes, that is already done :)

aplanas (approved):
Sorry that this takes so long : (

And thanks!

TobiPeterG:
> Sorry that this takes so long : (
> 
> And thanks!

All good :)

PR #63, opened 2024-03-05
Unified Kernel Image

val4oss:
# Goal

The aim of this PR, is to open discussion about the UKI feature for the sdbootutil tool.

# Implementation

- The idea would be to add two commands “add-uki” and “remove-uki” to add/remove an entry to boot in a potential Unified Kernel Image.
- Efi should be install into `/usr/lib/modules/$(uname)/<version format specification>.efi` as described at https://uapi-group.org/specifications/specs/unified_kernel_image/#distribution-built-ukis-installed-by-package-managers. I used the argument `--image`  to identified the uki name.
- UKIs in esp/loader/entries are automatically sourced by systemd-boot, FMHO it looks better to keep a loader configuration file to add an entry and remain homogeneous.

#  Corrections
- The approach may change, as **install_uki** do things as **install_kernel** did, a better way would be to create a function **install_image** called by the two previous functions.  What differs from UKI to a Kernel, is that it doesn’t install an initrd (as contained into the efi file) and the content of the entry conf file.

# Notes
- For now there isn’t any interaction with snapshots. I don’t yet get a solution to bring snapshots process with UKI as the CMDLINE is hidden into the binary and cannot be easily changed without break the signature. I am working on at, but the changes could be a good start to add uki entries to bootloader.

# Tests
Tested on Tumbleweed booting with sd-boot.

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -84,6 +84,12 @@ helpandquit()
 		bootloader [SNAPSHOT]
 			   Print the detected bootloader
 
+		add-uki VERSION [SNAPSHOT]
+			   Create boot entry for uki with specified kernel version
+
+		remove-uki VERSION [SNAPSHOT]
+			   Remove boot entry for uki with specified kernel version
+
 		add-kernel VERSION [SNAPSHOT]
 			   Create boot entry for specified kernel
 
@@ -658,6 +664,75 @@ make_free_space()
 
 	free_space="$(boot_free_space)"
 	[ "$total_size" -lt "$free_space" ]
+}
+
+install_uki() 
+{
+	local subvol="$1"
+	local kernel_version="$2" # Same as kernel version
+	local src="${subvol#${subvol_prefix}}/lib/modules/$kernel_version/uki.efi"
+	test -e "$src" || err "Can't find $src"
+
+	# According uapi group, uki name should contain version.
+    local uki_version=
+	uki_version="$(echo "$src" \
+	| sed 's/.*-\([0-9\.][0-9\.]*\).*/\1/' \
+	| sed 's/\.$//')"
+	if [ -z "$uki_version" ]; then
+	    warn "No version found for uki $image"
+	fi
+
+	calc_chksum "$src"
+	local dst="/$entry_token/$kernel_version/uki-$chksum"
+
+	# XXX: fix calling with snapshot instead of subvol
+	local snapshot="${subvol#${subvol_prefix}/.snapshots/}"
+	snapshot="${snapshot%/snapshot}"
+
+	mkdir -p "$boot_root${dst%/*}"
+
+	title="${os_release_PRETTY_NAME:-Linux $kernel_version}"
+	[ -z "$uki_version" ] || title="$title (UKI $uki_version)"
+	[ -n "$os_release_VERSION" ] || title="$title $os_release_VERSION_ID"
+
+	if ! is_transactional  && subvol_is_ro "$subvol"; then
+		set_snapper_title_and_sortkey "$snapshot"
+	fi
+
+	local entry_machine_id=
+	[ "$entry_token" = "$machine_id" ] && entry_machine_id="$machine_id"
+
+	cat > "$tmpdir/entry.conf" <<-EOF
+	# Boot Loader Specification type#1 entry
+	title      $title
+	efi        $dst
+	version    $snapshot@${uki_version+$uki_version-k}$kernel_version
+	EOF
+
+	local failed=
+	if [ ! -e "$boot_root$dst" ]; then
+		install_with_rollback "$src" "$boot_root$dst" || failed=kernel
+	fi
+	if [ -z "$failed" ]; then
+		loader_entry="$boot_root/loader/entries/$entry_token-uki-$kernel_version-$snapshot.conf"
+		install_with_rollback "$tmpdir/entry.conf" "$loader_entry" || failed="bootloader entry"
+	fi
+
+	rm -f "$tmpdir/entry.conf"
+	[ -z "$failed" ] || err "Failed to install $failed"
+	reset_rollback
+
+	# This action will require to update the PCR predictions
+	update_predictions=1
+}
+
+remove_uki() {
+	local subvol="$1"
+	local kernel_version="$2"
+	local snapshot="${subvol#${subvol_prefix}/.snapshots/}"
+	snapshot="${snapshot%/*}"
+	local id="$entry_token-uki-$kernel_version-$snapshot.conf"
+	run_command_output bootctl unlink "$id"
 }
 
 install_kernel()
@@ -2044,7 +2119,7 @@ if [ -z "$SYSTEMD_LOG_LEVEL" -a -n "$verbose" ]; then
 fi
 
 case "$1" in
-	install|needs-update|update|force-update|add-kernel|remove-kernel|set-default-snapshot|add-all-kernels|mkinitrd|remove-all-kernels|is-installed|list-snapshots|list-entries|list-kernels|show-entry|is-bootable|update-predictions|bootloader) ;;
+	install|needs-update|update|force-update|add-uki|remove-uki|add-kernel|remove-kernel|set-default-snapshot|add-all-kernels|mkinitrd|remove-all-kernels|is-installed|list-snapshots|list-entries|list-kernels|show-entry|is-bootable|update-predictions|bootloader) ;;
 	kernels|snapshots|entries|"") stty_size; interactive=1 ;;
 	*) err "unknown command $1" ;;
 esac
@@ -2108,6 +2183,10 @@ elif [ "$1" = "force-update" ]; then
 	if is_installed; then install_bootloader "${2:-$root_snapshot}"; else :; fi
 elif [ "$1" = "bootloader" ]; then
 	bootloader_name "${2:-$root_snapshot}"
+elif [ "$1" = "add-uki" ]; then
+	install_uki "${3:-$root_subvol}" "$2"
+elif [ "$1" = "remove-uki" ]; then
+	remove_uki "${3:-$root_subvol}" "$2"
 elif [ "$1" = "add-kernel" ]; then
 	install_kernel "${3:-$root_snapshot}" "$2"
 elif [ "$1" = "add-all-kernels" ]; then
--- a/sdbootutil
+++ b/sdbootutil
@@ -671,6 +671,9 @@ install_uki()
 	local subvol="$1"
 	local kernel_version="$2" # Same as kernel version
 	local src="${subvol#${subvol_prefix}}/lib/modules/$kernel_version/uki.efi"
+    if [ "$image" = "vmlinuz" ] || [ "$image" = "Image" ]; then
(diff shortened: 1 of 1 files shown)


aplanas:
>  I don’t yet get a solution to bring snapshots process with UKI as the CMDLINE is hidden into the binary and cannot be easily changed without break the signature. I am working on at

IIRC systemd has some proposals for that? I do not remember the details but was something about storing the cmdline as an external file

keentux:
> > I don’t yet get a solution to bring snapshots process with UKI as the CMDLINE is hidden into the binary and cannot be easily changed without break the signature. I am working on at
> 
> IIRC systemd has some proposals for that? I do not remember the details but was something about storing the cmdline as an external file

Indeed, I can use addons files to override the cmdline.
I didn't explain well my concern, sorry, but the problem is more about the signing process in the distribution provider point of view. If you need to have all files signed during your boot process, each addons (file) that override a section of your efi need to be signed, UKI can be signed by the distribution when the provider gives the image. But snapshots happens in runtime, therefore if you create an addon after the snapshot creation you cannot have a good signature. That why I am searching a way to have multi entries for an UKI according the number of snapshot without using a companion file. But this is maybe for another discussion topic :)

aplanas:
>  the problem is more about the signing process in the distribution provider point of view

Can the UKI be signed with and empty .cmdline PE section? If that is the case, the cmdline from the addon can be generated locally and delegated to the measure boot process instead of the secure boot one.

> That why I am searching a way to have multi entries for an UKI according the number of snapshot without using a companion file.

How can this work? the cmdline requires local information (root UUID, subvol path with the subvol number, etc)

aplanas:
Ah right!

There can be some objections to root uuid auto-discovering in the context of multiple OS installations.

But still curious about the rootflag=subvol= parameter[1], the combination of initrd + kernel is only valid for a subset of snapshots (and is the cmdline the one that links them). Do you think that there is a solution for this too?

[1] https://github.com/openSUSE/sdbootutil/blob/main/ARCHITECTURE.md#introducing-snapshots

keentux:
> Ah right!
> 
> There can be some objections to root uuid auto-discovering in the context of multiple OS installations.

Yes you are right, this methods is not adapted for multiple OS, or you should to have as UKI as root fs, but that is not something we may want :)

> 
> But still curious about the rootflag=subvol= parameter[1], the combination of initrd + kernel is only valid for a subset of snapshots (and is the cmdline the one that links them). Do you think that there is a solution for this too?
> 
> [1] https://github.com/openSUSE/sdbootutil/blob/main/ARCHITECTURE.md#introducing-snapshots

Unfortunately, I do not have any ideas yet about the rootflags parameter.. BTW there is no more combination Kernel+Initrd with the UKI as both are included in the one file. To provide the subvol info, I don't know how the bootloader could pass the information (because the bootloader already know the snapshots number as it is specified into the loader conf file in the Version parameter) to the UKI. I need to study more how snapshots process is done during the boot time

aplanas:
> > Do you mean in the `default` parameter? I do not see a `version` parameter in [1]
> 
> I mean Version from the loader conf file: `version 15@1.2.3-1-default`

Ah ... I cannot find it ... neither in the documentation nor in my systemd. 

> But you still can create an entry loader conf file and giving the path to an UKI in `esp/loader/entries/my-uki.conf`

Indeed ... like a chain load. I like this idea.

keentux:
> In theory the snapshot to boot could be extracted from the selected boot entry for a start. It's stored in the `4a67b082-0a4c-41cf-b6c7-440b29bb8c4f-LoaderEntrySelected` efi variable. So the initrd could parse it. Not the most elegant solution but could get us started.

Could be good to give a try for a first start yes.

>There was some discussion or pr somewhere about adding wildcard matching to kernel command lines for uki. I think it stalled and I can't find it right now though :/ that could also help us with the snapshot passing.

I found this discussion from systemd project, and talk about snapshots begins here : https://github.com/systemd/systemd/issues/24539#issuecomment-1508527353 maybe it is the PR you refereed for ?

lnussel:
missed the question part. yes that is the discussion

sdbootutil
```diff
@@ -660,6 +666,78 @@ make_free_space()
...
+}
+
+remove_uki() {
+	local subvol="$1"
+	local kernel_version="$2"
+	local snapshot="${subvol#${subvol_prefix}/.snapshots/}"
+	snapshot="${snapshot%/*}"
+	local id="$entry_token-uki-$kernel_version-$snapshot.conf"
+	run_command_output bootctl unlink "$id"
+}
```
aplanas, line 739: this is mostly the same code than remove_kernel (missing the prediction stuff), so I am wondering if would make sense to unify {add,remove}_kernel with {add,remove}_uki.

I am not sure that makes total sense, tho.

sdbootutil
```diff
@@ -2077,11 +2155,13 @@ fi
 [ -n "$root_subvol" ] || [ -z "$have_snapshots" ] || err "Can't determine root subvolume"
 [ -n "$root_device" ] || err "Can't determine root device"
 [ -n "$firmware_arch" ] || err "Can't determine firmware arch"
-case "$firmware_arch" in
-	x64) image=vmlinuz ;;
-	aa64) image=Image ;;
-	*) err "Unsupported architecture $firmware_arch" ;;
-esac
+if [ -z "$image" ]; then
```
aplanas, line 2158: why this?

sdbootutil
```diff
@@ -660,6 +666,78 @@ make_free_space()
...
+    local uki_version=
+	uki_version="$(echo "$src" \
+	| sed 's/.*-\([0-9\.][0-9\.]*\).*/\1/' \
+	| sed 's/\.$//')"
+	if [ -z "$uki_version" ]; then
+	    warn "No version found for uki $image"
+	fi
+
+	calc_chksum "$src"
+	local dst="/$entry_token/$kernel_version/uki-$chksum"
```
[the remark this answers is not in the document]
keentux: Not sure yet but probably not. There is a concept with "extensions", others efi binaries that can be load with the UKI to extend his initrd. In that way, we will have the same structur of kernel+initrd, so the hash could be useful in that case. I need to think about that

sdbootutil
keentux: It clearly makes sense to merge `$command_kernel` with `$command_uki`, and maybe rename the functions into `$command_image`.
First goal was to introduce the uki process, that why I split it in these new functions

sdbootutil
keentux: unwanted sorry

TobiPeterG:
A UKI can now contain multiple profiles, allowing to configure multiple, separate command lines.
This could be used, also shown in the linked discussion :)

keentux:
> A UKI can now contain multiple profiles, allowing to configure multiple, separate command lines. This could be used, also shown in the linked discussion :)

Profile feature is great to have multiple entries for the same UKI but with different command line. However, if I well understand profiles, these sections need to be included into the efi binary, for each new snapshot created, a called to ukify is needed and the efi binary will be regenerated, where in the vendor POV we signed the binary, it is not feasible.

However, as you suggested into the linked discussion, a multi-profiles with addons would works for snapshots. It will be closed of what I am testing with UKI and snapshots. I have wrote a short documentation on it : https://github.com/keentux/unified-kernel-image-tool/blob/main/docs/unified-kernel-image.md#v---snapshots

keentux:
Hi @TobiPeterG , Happy to see there are some interests of UKIs on openSUSE. :)

Indeed, there are some packages in Factory, available for Tumbleweed:
  1. **uki-tool**: a tool to regroup useful commands dealing with the UKI and adapted for the packaging.
      Sources: https://github.com/keentux/unified-kernel-image-tool
  2. **static-initrd-generic-unsigned**: provides Unsigned static initrd. (Package name may move for next update to include kernel flavors used)
  3. **uki-$kernelflavor**: provides Signed Unified Kernel Images. based on the kernel flavor (for now, only the default flavor is provided, will comme soon, the vanilla flavor) and based on the unsigned static initrd.
  
When installing the uki package it will automatically add a bootloader entry depending if you used grub2 or sdboot. I will recommend to not install yet the uki as I am working of an update that will improve the way bootloader entries are created. It is done by the uki-tool and the release should start asap.
  
You can find more documentation about static-initrd and uki here: https://en.opensuse.org/SDB:Static_Initrd_and_Unified_Kernel_Image
And finally, I made a short presentation last year (a bit old therefor) where the replay can be found here: https://www.youtube.com/watch?v=w1nMenPL1Uw

About this PullRequest status: I am working on the PR on systemd which I based my POC on to manage snapshots in openSUSE with the UKI, because with the new systemd version, my POC is broken and requires some changes.

I hope this provides some insight into the status of the UKI. Please feel free to contact me if you need any information or assistance. :grin:

TobiPeterG:
> Hi @TobiPeterG , Happy to see there are some interests of UKIs on openSUSE. :)
> 
> Indeed, there are some packages in Factory, available for Tumbleweed:
> 
> 1. **uki-tool**: a tool to regroup useful commands dealing with the UKI and adapted for the packaging.
>    Sources: https://github.com/keentux/unified-kernel-image-tool
> 2. **static-initrd-generic-unsigned**: provides Unsigned static initrd. (Package name may move for next update to include kernel flavors used)
> 3. **uki-$kernelflavor**: provides Signed Unified Kernel Images. based on the kernel flavor (for now, only the default flavor is provided, will comme soon, the vanilla flavor) and based on the unsigned static initrd.
> 
> When installing the uki package it will automatically add a bootloader entry depending if you used grub2 or sdboot. I will recommend to not install yet the uki as I am working of an update that will improve the way bootloader entries are created. It is done by the uki-tool and the release should start asap.
> 
> You can find more documentation about static-initrd and uki here: https://en.opensuse.org/SDB:Static_Initrd_and_Unified_Kernel_Image And finally, I made a short presentation last year (a bit old therefor) where the replay can be found here: https://www.youtube.com/watch?v=w1nMenPL1Uw
> 
> About this PullRequest status: I am working on the PR on systemd which I based my POC on to manage snapshots in openSUSE with the UKI, because with the new systemd version, my POC is broken and requires some changes.
> 
> I hope this provides some insight into the status of the UKI. Please feel free to contact me if you need any information or assistance. 😁

Thank you very much for this detailed report. :)
Can you link the systemd PR please? I had a look at
https://github.com/systemd/systemd/compare/main...keentux:systemd:addon-stanza-type1
Do you mean this branch? This would add a property to type 1 boot entries to select what addons should be loaded, right? I'm not sure if Lennard will like this approach, he's a fan of having these things automated.
I guess it would be logical (and great to automate) if you could only select addons when they override a property set in the UKI or two addons want to set the same property. This way there is no need to extend type 1 entries. :)

I would love to help with snapshot support. If you don't want/need help on that side for now, I would also love to work on the initrd and signing side. We need a way to add the modules the user wants/needs as addons to the base initrd. Do you know if dracut supports creating the addons?
I guess sdbootutil also needs support 
As for signing: I guess it should be possible to sign the addons locally uisng MOK to be able to boot them, right? Is that already supported?
Is there any other area you want/need support in? :)

aplanas:
Any news on this? I think UKIs are interesting, maybe we can retake it

PR #64, opened 2024-03-07
Fix update_kernel function

TobiPeterG:
c990b0f509f6e6fef3432f6b7e7abfb471d3d9c9 introduced a bug where all kernels where marked as stale. This fixes it.

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -710,7 +710,7 @@ list_snapshots()
 		if [ "$isdefault" = "true" ]; then
 			id="\e[1;4m$id\e[m"
 		fi
-		update_kernels "${subvol_prefix}/.snapshots/$n/snapshot"
+		update_kernels "$n"
 		[ "$is_bootable" = 1 ] || id="!$id"
 		echo -e "$id $title"
 	done < <(jq '.root|.[]|[.number, .default, .description]|join(" ")' -r < "$snapperfile")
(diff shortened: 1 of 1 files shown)


lnussel:
we need a test suite :)

TobiPeterG:
> we need a test suite :)

Thought about that as well.
I found https://github.com/bats-core/bats-core, but didn't invest too much time.
If wanted, I could try to setup some basic tests

TobiPeterG:
> > > we need a test suite :)
> > 
> > 
> > Thought about that as well. I found https://github.com/bats-core/bats-core, but didn't invest too much time. If wanted, I could try to setup some basic tests
> 
> How about a rewrite in a more settle language instead going into this rabbit hole? I propose this idea because a shell script with >1500 lines is something that worries me a lot.

I haven't used Rust yet, but would like to learn it, seems pretty neat.
Any objections? :P

This would, however, be quite time consuming, no matter which language is chosen

aplanas:
> I haven't used Rust yet, but would like to learn it, seems pretty neat. Any objections? :P

I am OK with this

> This would, however, be quite time consuming, no matter which language is chosen

Yes, we always do the same mistake: "nah this will be a small script to copy two files and iterate over some snapshots, will be alive until upstream gain some features and we can kill it later", but now is far for trivial and there are a bunch of features that I do not see fit upstream, so it will be alive for long.

wrt to the consumed time, testing it will be not a minor task. It will requires a lot of refactoring, as the functions are not ready to be tested (they read the fs, call tools that produce complex output, etc). So both path (rewrite + test, refactor + test) will be time consuming

PR #66, opened 2024-03-10
Make sdbootutil usable with no-snapshot root filesystems

didopl:
As sdboot is in some places a good alternative to grub as it's simpler and lighter, it would be nice if not only btrfs systems with snapper can use this, but also older systems with ext4 or btrfs where snapper is not installed and setup. Currently, sdboot can only be installed on btrfs, snapper systems with this tool. As this tool simplifies a lot when it comes to the install and update process of sdboot it would be convenient to use this on non snapshot systems.
Also from my experience the OS Installer fails to install sdboot when not using snapshots.

Note: This PR is not complete. A lot of functions are still missing and the current changes are so called "quick and dirty" and there might be a way to simplify this.

At current state of this PR, sdbootutil should successfully installs sdboot and the kernels.
The auto-detection should work now, so no more option is needed.

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -19,6 +19,7 @@ arg_all_entries=
 arg_no_variables=
 arg_no_reuse_initrd=
 arg_no_random_seed=
+have_snapshots=
 # for x in vmlinuz image vmlinux linux bzImage uImage Image zImage; do
 image=
 
@@ -234,6 +235,7 @@ is_transactional()
 
 subvol_is_ro()
 {
+	[ -n "$have_snapshots" ] || return 0
 	local subvol="${1:?}"
 	while read -r line; do
 		[ "$line" = "ro=true" ] && return 0
@@ -243,6 +245,7 @@ subvol_is_ro()
 
 detect_parent() {
 	local subvol="$1"
+	[ -n "$have_snapshots" ] || return 0
 	parent_uuid="$(btrfs subvol show "${subvol#"${subvol_prefix}"}" | sed -ne 's/\s*Parent UUID:\s*//p')"
 	[ "$parent_uuid" != '-' ] || parent_uuid=
 	[ -n "$parent_uuid" ] || return 0
@@ -257,7 +260,7 @@ detect_parent() {
 
 sedrootflags()
 {
-	local subvol="${1:?}"
+	local subvol="$1"
 	# - delete BOOT_IMAGE= and initrd=
 	# - make sure root= refers to uuid
 	# - replace or add rootflags to point at correct subvolume
@@ -273,21 +276,19 @@ sedrootflags()
 	# already there). Since we always operate on the same line,
 	# "empty" t jumps are used to reset the condition after very
 	# s///.
-	sed -e "s/[ \t]\+/ /g" \
-		-e "s/\<\(BOOT_IMAGE\|initrd\)=[^ ]* \?//" \
-		-e "ta;:a" \
-		-e "s/\<root=[^ ]*/root=UUID=$root_uuid/;tb;s,\$, root=UUID=$root_uuid,;tc;:c;:b" \
-		-e "s,\<rootflags=subvol=[^ ]*,rootflags=subvol=$subvol,;td;s,\$, rootflags=subvol=$subvol,;te;:e;:d" \
-		-e "s,\<systemd.machine_id=[^ ]*,systemd.machine_id=$machine_id,;tf;s,\$, systemd.machine_id=$machine_id,;tg;:g;:f"
-
+	local sed_arguments=("-e s/[ \t]\+/ /g"\
+		"-e s/\<\(BOOT_IMAGE\|initrd\)=[^ ]* \?//"\
+		"-e s/\<root=[^ ]*/root=UUID=$root_uuid/;tb;s,\$, root=UUID=$root_uuid,;tc;:c;:b"\
+		"-e s,\<systemd.machine_id=[^ ]*,systemd.machine_id=$machine_id,;tf;s,\$, systemd.machine_id=$machine_id,;tg;:g;:f")
+	[ -z "$have_snapshots" ] || sed_arguments+=("-e s,\<rootflags=subvol=[^ ]*,rootflags=subvol=$subvol,;td;s,\$, rootflags=subvol=$subvol,;te;:e;:d")
+	sed "${sed_arguments[@]}"
 }
 
 remove_kernel()
 {
-	local snapshot="${1:?}"
-	local subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
+	local snapshot="$1"
 	local kernel_version="$2"
-	local id="$entry_token-$kernel_version-$snapshot.conf"
+	local id="$entry_token-$kernel_version${snapshot:+-$snapshot}.conf"
 	run_command_output bootctl unlink "$id"
 
 	# This action will require to update the PCR predictions
@@ -319,6 +320,7 @@ update_snapper()
 
 set_snapper_title_and_sortkey()
 {
+	[ -n "$have_snapshots" ] || return 0
 	snapshot="${1:?}"
 	local type date desc important pre_num
 	local snapshot_info
@@ -345,8 +347,8 @@ set_snapper_title_and_sortkey()
 }
 
 reuse_initrd() {
-	local snapshot="${1:?}"
-	local subvol="${2:?}"
+	local snapshot="$1"
+	local subvol="$2"
 
 	[ -z "$arg_no_reuse_initrd" ] || return 1
 
@@ -413,13 +415,14 @@ add_version_to_title()
 
 install_kernel()
 {
-	local snapshot="${1:?}"
-	local subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
+	local snapshot="$1"
+	local subvol=""
+	[ -z "$have_snapshots" ] || subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
 	local kernel_version="$2"
 	local dstinitrd=()
 	local src="${subvol#"${subvol_prefix}"}/lib/modules/$kernel_version/$image"
 	local initrddir="${subvol#"${subvol_prefix}"}/usr/lib/initrd"
-	test -e "$src" || err "Can't find $src"
+	[ -e "$src" ] || err "Can't find $src"
 
 	calc_chksum "$src"
 	local dst="/$entry_token/$kernel_version/linux-$chksum"
@@ -441,8 +444,9 @@ install_kernel()
 	elif ! reuse_initrd "$snapshot" "$subvol"; then
 		local snapshot_dir="/.snapshots/$snapshot/snapshot"
 		local dracut_args=()
-		if [ "$subvol" != "$root_subvol" ]; then
-			dracut_args=('--sysroot' "${snapshot_dir}" '--tmpdir' '/var/tmp' '--add-device' "$root_device")
+		dracut_args=('--force' '--tmpdir' '/var/tmp' '--add-device' "$root_device")
+		if [ "$subvol" != "$root_subvol" ] && [ -n "$have_snapshots" ]; then
+			dracut_args+=('--sysroot' "${snapshot_dir}")
 		fi
 		log_info "generating new initrd"
 
@@ -514,7 +518,7 @@ install_kernel()
 		done
 	fi
 	if [ -z "$failed" ]; then
-		loader_entry="$boot_root/loader/entries/$entry_token-$kernel_version-$snapshot.conf"
+		loader_entry="$boot_root/loader/entries/$entry_token-$kernel_version${snapshot:+-$snapshot}.conf"
 		install_with_rollback "$tmpdir/entry.conf" "$loader_entry" || failed="bootloader entry"
 		rm -f "$tmpdir/entry.conf"
 	fi
@@ -527,7 +531,7 @@ install_kernel()
 
 install_all_kernels()
 {
-	local snapshot="${1:?}"
+	local snapshot="$1"
 	find_kernels "$snapshot"
 	for kv in "${!found_kernels[@]}"; do
 		log_info "installing $kv"
@@ -538,7 +542,7 @@ install_all_kernels()
 
 remove_all_kernels()
 {
-	local snapshot="${1:?}"
+	local snapshot="$1"
 	find_kernels "$snapshot"
 	for kv in "${!found_kernels[@]}"; do
 		remove_kernel "${snapshot}" "$kv"
@@ -604,6 +608,7 @@ list_entries()
 						errors+=("$root/$v does not exist")
 					fi
 				fi
+				[ -n "$have_snapshots" ] || break
 				if [ "$k" = 'options' ]; then
 					local snapshot
 					read -r snapshot <<<"$(echo "$v" | sed -e "s,.*rootflags=subvol=${subvol_prefix}/.snapshots/\([0-9]\+\)/snapshot.*,\1,")"
@@ -739,6 +744,7 @@ show_entry()
 
 list_snapshots()
 {
+	[ -n "$have_snapshots"  ] || { log_info "System does no support snapshots."; return 0; }
 	update_snapper 2>"$tmpfile" || err "$(cat "$tmpfile")"
 
 	local n=0
@@ -756,6 +762,7 @@ list_snapshots()
 
 show_snapper()
 {
+	[ -n "$have_snapshots" ] || { log_info "System does no support snapshots."; return 0; }
 	if ! update_snapper 2>"$tmpfile"; then
 		d --title "Error" --textbox "$tmpfile" 0 0
 		exit 1
@@ -819,7 +826,8 @@ calc_chksum() {
 declare -A found_kernels
 find_kernels()
 {
-	local subvol="${subvol_prefix}/.snapshots/${1:?}/snapshot"
+	local subvol=""
+	[ -z "$have_snapshots" ] || subvol="${subvol_prefix}/.snapshots/${1:?}/snapshot"
 	local fn kv
 	found_kernels=()
 
@@ -840,7 +848,7 @@ declare -A stale_kernels
 is_bootable=
 update_kernels()
 {
-	local snapshot="${1:?}"
+	local snapshot="$1"
 	local path id
 	installed_kernels=()
 	stale_kernels=()
@@ -865,7 +873,8 @@ update_kernels()
 
 list_kernels()
 {
-	local snapshot="${1:?}"
+	local snapshot=""
+	[ -z "$have_snapshots" ] || snapshot="${1:?}"
 	update_kernels "$snapshot"
 	local kernelfiles=("${!installed_kernels[@]}")
 	for k in "${kernelfiles[@]}"; do
@@ -887,7 +896,7 @@ list_kernels()
 
 is_bootable()
 {
-	local snapshot="${1:?}"
+	local snapshot="$1"
 	update_kernels "$snapshot"
 
 	[ "$is_bootable" = 1 ] || return 1
@@ -896,8 +905,9 @@ is_bootable()
 
 show_kernels()
 {
-	local snapshot="${1:?}"
-	subvol="${subvol_prefix}/.snapshots/${1:?}/snapshot"
+	local subvol=""
+	local snapshot="$1"
+	[ -z "$have_snapshots" ] || subvol="${subvol_prefix}/.snapshots/${1:?}/snapshot"
 	while true; do
 		update_kernels "$snapshot"
 		local list=()
@@ -997,7 +1007,8 @@ is_installed()
 
 find_sdboot()
 {
-	local prefix="/.snapshots/${1-$root_snapshot}/snapshot"
+	local prefix=""
+	[ -z "$have_snapshots" ] || prefix="/.snapshots/${1-$root_snapshot}/snapshot"
 	# XXX: this is a hack in case we need to inject a signed
 	# systemd-boot from a separate package
 	local sdboot="$prefix/usr/lib/systemd-boot/systemd-boot$firmware_arch.efi"
@@ -1007,7 +1018,8 @@ find_sdboot()
 
 find_grub2()
 {
-	local prefix="/.snapshots/${1-$root_snapshot}/snapshot"
+	local prefix=""
+	[ -z "$have_snapshots" ] || prefix="/.snapshots/${1-$root_snapshot}/snapshot"
 	local grub2="$prefix/usr/share/efi/$(uname -m)/grub.efi"
 	[ -e "$grub2" ] || grub2="$prefix/usr/share/grub2/$(uname -m)-efi/grub.efi"
 	echo "$grub2"
@@ -1026,8 +1038,12 @@ find_bootloader()
 
 bootloader_needs_update()
 {
-	local snapshot="${1-$root_snapshot}"
-	local prefix="/.snapshots/${snapshot}/snapshot"
+	local prefix=""
+	local snapshot=""
+	if [ -n "$have_snapshots" ]; then
+		snapshot="${1-$root_snapshot}"
+		prefix="/.snapshots/${snapshot}/snapshot"
+	fi
 	local bldr_name
 	local v nv
 	v="$(bootloader_version)"
@@ -1044,8 +1060,12 @@ bootloader_needs_update()
 
 install_bootloader()
 {
-	local snapshot="${1:-$root_snapshot}"
-	local prefix="/.snapshots/${root_snapshot}/snapshot"
+	local snapshot=""
+	local prefix=""
+	if [ -n "$have_snapshots" ]; then
+		snapshot="${1:-$root_snapshot}"
+		prefix="/.snapshots/${root_snapshot}/snapshot"
+	fi
 	local bootloader bldr_name blkpart drive partno
 
 	bootloader=$(find_bootloader "$snapshot")
@@ -1208,6 +1228,7 @@ set_default_entry()
 
 set_default_snapshot()
 {
+	[ -n "$have_snapshots" ] || { log_info "System does no support snapshots."; return 0; }
 	local num="${1:?}"
 	local configs
 	update_entries_for_snapshot "$num"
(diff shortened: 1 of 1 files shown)


sdbootutil
```diff
@@ -20,6 +20,7 @@ arg_all_entries=
 arg_no_variables=
 arg_no_reuse_initrd=
 arg_no_random_seed=
+no_snapshots=0
```
lnussel, line 23: maybe call it have_snapshots

sdbootutil
```diff
@@ -63,6 +64,7 @@ helpandquit()
 		  --image		Specify Linux kernel file name
 		  --no-variables	Do not update UEFI variables
 		  --no-reuse-initrd	Always regenerate initrd
+		  --no-snapshots	The running system does not support snapshots
```
lnussel, line 67: I think this should be auto detected instead

sdbootutil
```diff
@@ -214,6 +216,9 @@ is_transactional()
 
 subvol_is_ro()
 {
+	if [ $no_snapshots -eq 1 ]; then
+		return 0
+	fi
```
lnussel, line 221: would turn into something like
`[ -n "$have_snapshots" ] || return 0`

sdbootutil
```diff
@@ -223,7 +228,11 @@ subvol_is_ro()
 
 detect_parent() {
 	local subvol="$1"
-	parent_uuid="$(btrfs subvol show "${subvol#${subvol_prefix}}" | sed -ne 's/\s*Parent UUID:\s*//p')"
+	if [ $no_snapshots -eq 1 ]; then
+		parent_uuid=""
+	else
+		parent_uuid="$(btrfs subvol show "${subvol#${subvol_prefix}}" | sed -ne 's/\s*Parent UUID:\s*//p')"
+	fi
```
lnussel, line 235: I guess the function could return early without snapshots

sdbootutil
```diff
@@ -237,7 +246,12 @@ detect_parent() {
 
 sedrootflags()
 {
-	local subvol="${1:?}"
+	if [ $no_snapshots -eq 1 ]; then
+		local subvol=""
+	sed -e "s/[ \t]\+/ /g;s/\<\(BOOT_IMAGE\|initrd\)=[^ ]* \?//;s/\<root=[^ ]*/root=UUID=$root_uuid/;tr;:r;:x"
+	else
+		local subvol="${1:?}"
+	fi
```
lnussel, line 254: maybe have the common part in a variable and add the subvol bits only if snapshots are enabled?

sdbootutil
```diff
@@ -246,10 +260,21 @@ sedrootflags()
...
+		snapshot=""
+		subvol=""
+	else
+		local snapshot="${1:?}"
+		local subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
+		snapshot="${snapshot%/*}"
+	fi
+	
 	local kernel_version="$2"
+	echo "kerbel_version in remove_kernel: $kernel_version"
```
lnussel, line 273: left over debugging code

sdbootutil
```diff
@@ -246,10 +260,21 @@ sedrootflags()
...
 remove_kernel()
 {
-	local snapshot="${1:?}"
-	local subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
+	if [ $no_snapshots -eq 1 ]; then
+		snapshot=""
+		subvol=""
+	else
+		local snapshot="${1:?}"
+		local subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
```
lnussel, line 268: subvol is not actually used in this function

sdbootutil
```diff
@@ -246,10 +260,21 @@ sedrootflags()
...
+		snapshot="${snapshot%/*}"
+	fi
+	
 	local kernel_version="$2"
+	echo "kerbel_version in remove_kernel: $kernel_version"
+	# local snapshot="${subvol#${subvol_prefix}/.snapshots/}"
+	
 	local id="$entry_token-$kernel_version-$snapshot.conf"
```
lnussel, line 276: that would leave a dash if snapshot is empty

sdbootutil
```diff
@@ -246,10 +260,21 @@ sedrootflags()
...
+	echo "id: $id"
```
lnussel, line 277: leftover debug

sdbootutil
```diff
@@ -306,8 +334,13 @@ set_snapper_title_and_sortkey()
...
 reuse_initrd() {
-	local snapshot="${1:?}"
-	local subvol="${2:?}"
+	if [ $no_snapshots -eq 1 ]; then
+		local snapshot=""
+		local subvol=""
+	else
+		local snapshot="${1:?}"
+		local subvol="${2:?}"
+	fi
```
lnussel, line 343: just remove the `:?` in such code I think. It's mostly an assert. Could add an extra function that checks for non-empty in case snapshots are enabled

sdbootutil
```diff
@@ -399,10 +443,17 @@ install_kernel()
...
 		done
+
 		/usr/bin/mkmoduleinitrd "${subvol#"${subvol_prefix}"}" "$kernel_version" "$tmpdir/initrd-$i"
+		
 	elif ! reuse_initrd "$snapshot" "$subvol"; then
 		local snapshot_dir="/.snapshots/$snapshot/snapshot"
 		local dracut_args=()
+		if [ $no_snapshots -eq 1 ]; then
+			snapshot_dir="/"
+			dracut_args=('--force' '--tmpdir' '/var/tmp' '--add-device' "$root_device")
```
lnussel, line 454: just store this common part first and only add the sysroot option if snapshots are enabled

sdbootutil
```diff
@@ -435,6 +486,8 @@ install_kernel()
 	fi
 
 	title="${os_release_PRETTY_NAME:-Linux $kernel_version}"
+	# TW pretty name does not include the version
+	[ -n "$os_release_VERSION" ] || title="$title $os_release_VERSION_ID"
```
lnussel, line 490: IIRC code related to the title was or will be refactored in a different pr. Just omit for now

sdbootutil
```diff
@@ -500,10 +557,16 @@ install_all_kernels()
...
 remove_all_kernels()
 {
-	local snapshot="${1:?}"
+	if [ $no_snapshots -eq 1 ]; then
+		local snapshot=""
+	else
+		local snapshot="${1:?}"
+	fi
 	find_kernels "$snapshot"
+	echo ${!found_kernels[@]}
```
lnussel, line 566: leftover debug

sdbootutil
```diff
@@ -567,10 +630,18 @@ list_entries()
...
+						read -r snapshot <<<"$(echo "$v" | sed -e "s,.*rootflags=subvol=${subvol_prefix}\([0-9]\+\),\1,")"
+						if [ -n "$v" ]; then
+							echo $snapshot
+						fi
+					else
+						local snapshot
+						read -r snapshot <<<"$(echo "$v" | sed -e "s,.*rootflags=subvol=${subvol_prefix}/.snapshots/\([0-9]\+\)/snapshot.*,\1,")"
+						if [ ! -d "/.snapshots/$snapshot/snapshot" ]; then
+							errors+=("/.snapshot/$snapshot/snapshot does not exist")
+						fi
```
lnussel, line 644: this whole block doesn't make sense without snapshots AFAICT

sdbootutil
```diff
@@ -701,6 +772,9 @@ show_entry()
 
 list_snapshots()
 {
+	if [ $no_snapshots -eq 1 ]; then
+		err "list-snapshots incompatible with --no-snapshots!"
+	fi
```
lnussel, line 777: just silently do nothing?

sdbootutil
```diff
@@ -1024,7 +1136,12 @@ install_sdboot()
 
 	[ -s "$boot_root/loader/entries.srel" ] || echo type1 > "$boot_root/loader/entries.srel"
 
-    	[ -e "$boot_root/loader/loader.conf" ] || echo -e "#timeout 3\n#console-mode keep\n" > "$boot_root/loader/loader.conf"
+	# echo "boot_root: $boot_root"
+	if [ $no_snapshots -eq 1 ]; then
+		echo -e "timeout 5\n#console-mode keep\n" > "$boot_root/loader/loader.conf"
+	else
+		[ -e "$boot_root/loader/loader.conf" ] || echo -e "#timeout 3\n#console-mode keep\n" > "$boot_root/loader/loader.conf"
+	fi
```
lnussel, line 1144: doesn't make sense

sdbootutil
```diff
@@ -1437,8 +1555,13 @@ done
...
 read -r root_uuid root_device < <(findmnt / -v -r -n -o UUID,SOURCE)
-root_subvol=$(btrfs subvol show / 2>/dev/null|head -1)
-subvol_prefix="${root_subvol%/.snapshots/*}"
+if [ $no_snapshots -eq 1 ]; then
+	root_subvol=""
+	sub_vol_prefix=""
+else
+	root_subvol=$(btrfs subvol show / 2>/dev/null|head -1)
+	subvol_prefix="${root_subvol%/.snapshots/*}"
+fi
```
lnussel, line 1564: this is where we can do the autodetection.
```
if [ "$(stat -f -c %T /)" = "btrfs" ] && [ -d /.snapshots ]; then
  have_snapshots=1
  root_subvol=...
```

lnussel (changes requested):

sdbootutil
didopl: Since snapshotting is getting more widely used, I would this maybe have_no_snapshots. This also requires less code changings.

sdbootutil
didopl: Agree.

sdbootutil
didopl: I'm not a big fan of doing things completely silent, but a warning might be better in this case.

sdbootutil
didopl: For me, I was unable to choose which entry to boot, since it didn't add a timeout in loader.conf

sdbootutil
didopl: Yep.

sdbootutil
lnussel: that's how bootctl install behaves by default so we keep that. it's not related to snapshots

sdbootutil
didopl: I'm not familiar at all with sed. Where does sed store the output? Is it written to a file as it seems this line sets the cmdline for the kernel or is it stored i an variable which is written to file then? I would appreciate some help here.

sdbootutil
didopl: So just leave out the `:?` and no if/test-statement?

sdbootutil
didopl: Yes.

sdbootutil
didopl: So when using have_no_snapshots instead, this would look like this:
`[ -n "$have_no_snapshots" ] && return 0`

sdbootutil
didopl: So just add `[ -n "$have_no_snapshots" ] && break` before?

sdbootutil
didopl: So to make it work with have_no_snapshots:
  
```
if ! { [ "$(stat -f -c %T /)" = "btrfs" ] && [ -d /.snapshots ]; }; then
    have_no_snapshots=1  
    root_subvol=""  
    sub_vol_prefix=""  
else...
```

sdbootutil
didopl: Ok, interesting. I will delete this.

sdbootutil
didopl: Added.

sdbootutil
didopl: Should be added now

sdbootutil
didopl: Is removed.

sdbootutil
didopl: Removed.

sdbootutil
didopl: Removed.

sdbootutil
didopl: Should `return 0` now.

sdbootutil
didopl: Auto detection works from my tests.

sdbootutil
didopl: Removed.

sdbootutil
didopl: Removed.

sdbootutil
didopl: I just do it in two commands now.

sdbootutil
didopl: I tested with a snapshot system and it installed correctly, so auto-detection should be fine.

sdbootutil
lnussel: the point of have_snapshots would be that we could potentially define other values in the future to support a different form of snapshots, not just the snapper way.

sdbootutil
lnussel: yes

sdbootutil
```diff
@@ -374,22 +397,31 @@ add_version_to_title()
...
 	local kernel_version="$2"
+	local src="${subvol#${subvol_prefix}}/lib/modules/$kernel_version/$image"
 	local initrd=()
 	local dstinitrd=()
-	local src="${subvol#${subvol_prefix}}/lib/modules/$kernel_version/$image"
-	local initrddir="${subvol#${subvol_prefix}}/usr/lib/initrd"
-	test -e "$src" || err "Can't find $src"
+	local initrddir=${subvol#${subvol_prefix}}/usr/lib/initrd
 
+	test -e "$src" || err "Can't find $src"
```
lnussel, line 413: Wonder how test creeped in here. Should be `[]`. Anyway the reordering here doesn't seem to be necessary?

sdbootutil
```diff
@@ -374,22 +397,31 @@ add_version_to_title()
...
+	
 	calc_chksum "$src"
 	local dst="/$entry_token/$kernel_version/linux-$chksum"
 
+	# XXX: fix calling with snapshot instead of subvol
+	local snapshot="${subvol#${subvol_prefix}/.snapshots/}"
+	snapshot="${snapshot%/snapshot}"
```
lnussel, line 420: broken rebase? This should not exist

sdbootutil
```diff
@@ -435,6 +472,8 @@ install_kernel()
 	fi
 
 	title="${os_release_PRETTY_NAME:-Linux $kernel_version}"
+	# TW pretty name does not include the version
+	# [ -n "$os_release_VERSION" ] || title="$title $os_release_VERSION_ID"
```
lnussel, line 476: unrelated, again probably result of rebase fallout

sdbootutil
```diff
@@ -476,7 +515,9 @@ install_kernel()
 	fi
 	if [ -z "$failed" ]; then
 		loader_entry="$boot_root/loader/entries/$entry_token-$kernel_version-$snapshot.conf"
+		[ -n "$have_no_snapshots" ] && loader_entry="$boot_root/loader/entries/$entry_token-$kernel_version.conf"
```
lnussel, line 518: `"...${snapshot:+-$snapshot}.conf"`

sdbootutil
```diff
@@ -567,6 +618,8 @@ list_entries()
 					fi
 				fi
 				if [ "$k" = 'options' ]; then
+					[ -n "$have_no_snapshots" ] && break
+
```
lnussel, line 622: skip parsing the options line one line above?

sdbootutil
```diff
@@ -951,7 +1024,11 @@ is_installed()
 
 find_sdboot()
 {
-	local prefix="/.snapshots/${1-$root_snapshot}/snapshot"
+	if [ -n "$have_no_snapshots" ]; then
+		local prefix=""
```
lnussel, line 1028: just move the declaration out and only set prefix if snapshots are enabled

sdbootutil
```diff
@@ -1024,7 +1113,9 @@ install_sdboot()
 
 	[ -s "$boot_root/loader/entries.srel" ] || echo type1 > "$boot_root/loader/entries.srel"
 
-    	[ -e "$boot_root/loader/loader.conf" ] || echo -e "#timeout 3\n#console-mode keep\n" > "$boot_root/loader/loader.conf"
+	# echo "boot_root: $boot_root"
+	[ -e "$boot_root/loader/loader.conf" ] || echo -e "#timeout 3\n#console-mode keep\n" > "$boot_root/loader/loader.conf"
```
lnussel, line 1117: not needed

sdbootutil
lnussel: no else needed here with `have_snapshots`

lnussel (changes requested):

sdbootutil
didopl: That makes totally sense of course. So I will try to do it with have_snapshots.

sdbootutil
didopl: Removed.

sdbootutil
didopl: Added.

sdbootutil
didopl: Removed.

sdbootutil
didopl: Using have_snapshots now.

sdbootutil
didopl: Removed.

sdbootutil
```diff
@@ -1550,7 +1596,7 @@ settle_entry_token
 
 [ -n "$boot_root" ] || err "No ESP detected. Legacy system?"
 [ -n "$root_uuid" ] || err "Can't determine root UUID"
-[ -n "$root_subvol" ] || err "Can't determine root subvolume"
+[ -n "$root_subvol" ] || [ -z "$have_snashots" ] || err "Can't determine root subvolume"
```
lnussel, line 1599: typo

sdbootutil
```diff
@@ -1538,8 +1579,13 @@ done
 # XXX: bootctl should have json output for that too
 eval "$(bootctl 2>/dev/null | sed -ne 's/Firmware Arch: *\(\w\+\)/firmware_arch="\1"/p;s/ *token: *\(\w\+\)/entry_token="\1"/p;s, *\$BOOT: *\([^ ]\+\).*,boot_root="\1",p')"
 read -r root_uuid root_device < <(findmnt / -v -r -n -o UUID,SOURCE)
-root_subvol=$(btrfs subvol show / 2>/dev/null|head -1)
-subvol_prefix="${root_subvol%/.snapshots/*}"
+root_subvol=""
+sub_vol_prefix=""
```
lnussel, line 1583: typo

sdbootutil
```diff
@@ -1208,6 +1248,7 @@ set_default_entry()
 
 set_default_snapshot()
 {
+	[ -n "$have_snapshots" ] || log_info "System does not support snaphsots." && exit 0
```
lnussel, line 1251: `[ -n "$have_snapshots" ] || { log_info ...; return 0; }`

Same in several other places wherever this pattern appears

sdbootutil
```diff
@@ -257,7 +260,11 @@ detect_parent() {
 
 sedrootflags()
 {
+	sed -e "s/[ \t]\+/ /g;s/\<\(BOOT_IMAGE\|initrd\)=[^ ]* \?//;s/\<root=[^ ]*/root=UUID=$root_uuid/;tr;:r;:x"
+	# echo "p1: ${1:?}"
+	[ -n "$have_snapshots" ] || return 0
 	local subvol="${1:?}"
+	sed -e "s,\<rootflags=subvol=[^ ]*,rootflags=subvol=$subvol,;tx; s,\$, rootflags=subvol=$subvol,;:x"
```
lnussel, line 267: please rebase. You can now move the sed arguments into an array and add the subvol part on demand

sdbootutil
```diff
@@ -284,10 +291,13 @@ sedrootflags()
 
 remove_kernel()
 {
-	local snapshot="${1:?}"
-	local subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
+	local snapshot="${1}"
+	[ -n "$have_snapshots" ] && snapshot="${snapshot%/*}"
```
lnussel, line 295: no longer needed

sdbootutil
```diff
@@ -345,8 +356,8 @@ set_snapper_title_and_sortkey()
 }
 
 reuse_initrd() {
-	local snapshot="${1:?}"
-	local subvol="${2:?}"
+	local snapshot="${1}"
+	local subvol="${2}"
```
lnussel, line 360: would look better without `{}`

sdbootutil
```diff
@@ -413,17 +424,22 @@ add_version_to_title()
 
 install_kernel()
 {
-	local snapshot="${1:?}"
-	local subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
+	local snapshot="${1}"
+	local subvol=""
+	[ -n "$have_snapshots" ] && subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
```
lnussel, line 429: `[ -z "$have_snapshots" ] || subvol=...`

sdbootutil
```diff
@@ -413,17 +424,22 @@ add_version_to_title()
...
-
+	[ -e "$src" ] || err "Can't find $src"
+	
 	calc_chksum "$src"
 	local dst="/$entry_token/$kernel_version/linux-$chksum"
 
+	# XXX: fix calling with snapshot instead of subvol
+	# local snapshot="${subvol#${subvol_prefix}/.snapshots/}"
+	# snapshot="${snapshot%/snapshot}"
+	# [ -n "$have_snapshots" ] || snapshot=""
```
lnussel, line 442: not needed

sdbootutil
```diff
@@ -460,6 +479,7 @@ install_kernel()
 	for i in /etc/kernel/cmdline /usr/lib/kernel/cmdline /proc/cmdline; do
 		[ -f "$i" ] || continue
 		boot_options="$(sedrootflags "$subvol" < "$i")"
+		echo $boot_options
```
lnussel, line 482: debug

sdbootutil
```diff
@@ -896,8 +926,8 @@ is_bootable()
 
 show_kernels()
 {
-	local snapshot="${1:?}"
-	subvol="${subvol_prefix}/.snapshots/${1:?}/snapshot"
+	subvol=""
+	[ -n "$have_snapshots" ] && subvol="${subvol_prefix}/.snapshots/${1:?}/snapshot"
```
lnussel, line 930: you've dropped the snapshot var

sdbootutil
```diff
@@ -413,17 +424,22 @@ add_version_to_title()
...
+	local subvol=""
+	[ -n "$have_snapshots" ] && subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
 	local kernel_version="$2"
 	local dstinitrd=()
 	local src="${subvol#"${subvol_prefix}"}/lib/modules/$kernel_version/$image"
 	local initrddir="${subvol#"${subvol_prefix}"}/usr/lib/initrd"
-	test -e "$src" || err "Can't find $src"
-
+	[ -e "$src" ] || err "Can't find $src"
+
```
lnussel, line 435: extra whitespace

lnussel (changes requested):
please use shellcheck, that would eliminate some trivial mistakes

sdbootutil
didopl: Changed in some other places too, not sure if necessary, but works.

sdbootutil
didopl: Removed.

sdbootutil
didopl: Works now from my tests.

sdbootutil
didopl: Changed.

sdbootutil
didopl: Removed.

sdbootutil
didopl: Removed.

sdbootutil
didopl: snapshot added again.

sdbootutil
didopl: Removed.

sdbootutil
didopl: See below.

sdbootutil
didopl: Should be added.

sdbootutil
didopl: Line is moved.

sdbootutil
didopl: Replaced with [].

sdbootutil
didopl: Fixed.

sdbootutil
didopl: Fixed.

sdbootutil
didopl: Replaced.

sdbootutil
```diff
@@ -896,8 +905,9 @@ is_bootable()
 
 show_kernels()
 {
-	local snapshot="${1:?}"
-	subvol="${subvol_prefix}/.snapshots/${1:?}/snapshot"
+	local subvol=""
+	local snapshot=""
+	[ -z "$have_snapshots" ] || subvol="${subvol_prefix}/.snapshots/${1:?}/snapshot"
```
lnussel, line 910: does not set snapshot anymore now

lnussel (changes requested):
those two minor things otherwise looks quite good

Please rebase and squash on main to have a single commit.

sdbootutil
```diff
@@ -865,7 +873,8 @@ update_kernels()
 
 list_kernels()
 {
-	local snapshot="${1:?}"
+	local snapshot=""
+	[ -z "$have_snapshots" ] || local snapshot="${1:?}"
```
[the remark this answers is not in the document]
didopl: Removed.

lnussel:
`git rebase -i` is quite nice as it allows you to reorder, squash and edit commits. Anyway good to go code wise! Use `git commit --amend --reset-author` to produce a descriptive commit message. You may also want to configure your git settings before to have a proper name and email address (`man git-config`).

lnussel (approved):

PR #67, opened 2024-03-11
Add initial GRUB2 support

aplanas:


sdbootutil
```diff
@@ -155,6 +155,25 @@ warn()
 	fi
 }
 
+is_sdboot()
+{
+	# If systemd-boot and grub2 are co-installed, we favor grub2
+	# in the detection
+	local sdboot grub2
+	sdboot="$(find_sdboot)"
+	grub2="$(find_grub2)"
```
lnussel, line 164: pass snapshot as argument? That way it would be possible to change the bootloader based on target snapshot

sdbootutil
```diff
@@ -269,7 +288,7 @@ install_with_rollback()
 		mv "$dst" "$dst.bak" || return "$?"
 	fi
 	rollback+=("$dst")
-	install -m 0644 "$src" "$dst" || return "$?"
+	install -p -m 0644 "$src" "$dst" || return "$?"
```
lnussel, line 291: does this work? FAT has lower precision time stamps IIIRC

sdbootutil
```diff
@@ -956,33 +983,54 @@ find_sdboot()
...
 	local sdboot="$prefix/usr/lib/systemd-boot/systemd-boot$firmware_arch.efi"
 	[ -e "$sdboot" ] || sdboot="$prefix/usr/lib/systemd/boot/efi/systemd-boot$firmware_arch.efi"
-	[ -e "$sdboot" ] || err "missing $sdboot"
 	echo "$sdboot"
 }
 
-sdboot_needs_update()
+find_grub2()
+{
+	local prefix="/.snapshots/${1-$root_snapshot}/snapshot"
```
lnussel, line 991: would do the fallback to root_snapshot in the caller. I think that's the way it's done in other functions

sdbootutil
```diff
@@ -995,36 +1043,39 @@ install_sdboot()
...
 	update_random_seed
 
-	[ -s "$boot_root/loader/entries.srel" ] || echo type1 > "$boot_root/loader/entries.srel"
-
-    	[ -e "$boot_root/loader/loader.conf" ] || echo -e "#timeout 3\n#console-mode keep\n" > "$boot_root/loader/loader.conf"
+	if is_sdboot; then
+		[ -s "$boot_root/loader/entries.srel" ] || echo type1 > "$boot_root/loader/entries.srel"
+		[ -e "$boot_root/loader/loader.conf" ] || echo -e "#timeout 3\n#console-mode keep\n" > "$boot_root/loader/loader.conf"
+	elif is_grub2; then
+		rsync -aL --delete --safe-links "$boot_root/../grub2" "$boot_root"
```
lnussel, line 1077: We need to sync from the original source in /usr here. /boot might be the XBOOTLDR partition in the future end esp not mounted in /boot/efi

Does grub.efi actually need all those files anyway? thought it was modules from a time before secure boot when grub had to have all modules built in?

lnussel (changes requested):

sdbootutil

sdbootutil
aplanas: The theme and fonts are there (and in /usr/share) but also the grub.cfg.

aplanas: removed for now, I am trying a different approach

sdbootutil
aplanas: uhm no, check find_sdboot: https://github.com/openSUSE/sdbootutil/blob/main/sdbootutil#L954

lnussel:
We probably need to bikeshed over the dialog title a bit otherwise looks fine

lnussel (approved):

aplanas:
I will keep it a bit longer in draft mode, as the last commit can change a bit: I want to generate the grub.cfg in the installation verb

aplanas:
Ready for full review. The last commit add a default grub.cfg and sync one single module (bli.mod) that is needed to the EFI vars

Issue #26, opened 2023-11-22
discussion: rethink out-of-space approach


TobiPeterG:
Currently, when there's not enough space on the ESP, the command exits with an error code. However, this might not always be obvious. This might lead to a system with no bootable entry for the current snapshot. The user then has to manually remove snapshots until old kernels got deleted and then recreate the entries to get a booting system with the current snapshot again.

Instead, one might choose a different approach to handling this issue:
If a kernel should be generated for the most recent/rw snapshot,
a) delete the oldest kernel(s) and its associated snapshots. The oldest kernel is probably used by the oldest snaphsots and therefore likely irrelevant
b) delete the kernel with the least number of associated snapshots. This way, the smallest possible number of snapshots are deleted

This way, the user autmatically has a booting system.

lnussel:
yes, we could also just remove the boot entries for old snapshots without actually cleaning up the snapshots. There's likely more space for snapshots than for kernel/initrds in the ESP

throttlemeister:
If you leave the snapshot, you can still recover files from it, but it is no longer bootable. I guess that is ok. What is more troublesome, is that snapper prunes old snapshots (good) but entries in the boot menu are not. It is therefor likely a lot more boot entries exist for non-existent snapshots than the other way around. See also my comment in the other issue.

lnussel:
when removing snapshots, snapper is supposed to call the sdbootutil plugin and remove kernels and entries from the ESP first. Do you by chance have logs of a case where this didn't happen?

lnussel:
The plugin is in `sdbootutil-snapper`, should be pulled in automatically. Snapper logs into /var/log/snapper.log

throttlemeister:
I checked the logs. I can see sdbootutil-snapper cleans up snapshot entries like pre and post, but no other snapshots. So there are still plenty snapshot configs in the bootloader related to kernels that are no longer present on the system.

Example:
# snapper list shows 19 snapshot entries listed for root
 
# sdbootutil kernels show 6.6.10 and 6.6.11 installed

And finally an ls of /boot/efi/loader/entries shows 54 snapshot configs for the bootmenu, the oldest being for kernel 6.6.3

lnussel:
Could have happened during development. The question whether it still happens today with current snapper and sdbootutil. That is unrelated to this ticket so I'd suggest to open a new one if you see new snapshots also causing stale files.

[...]

PR #70, opened 2024-03-12
Remove old kernels when no space is available

TobiPeterG:
This will clean up old kernels to free space on the ESP when a new kernel should be installed. Only kernels that are not installed in the to be installed snapshot and root snapshot (current snaphsot) might be removed

fixes https://github.com/openSUSE/sdbootutil/issues/26

Tested this on Tumbleweed

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -21,6 +21,8 @@ arg_no_variables=
 arg_no_reuse_initrd=
 arg_no_random_seed=
 have_snapshots=
+src_size_kb=
+declare -A pending_esp_files
 # for x in vmlinuz image vmlinux linux bzImage uImage Image zImage; do
 image=
 
@@ -330,6 +332,13 @@ update_entries_for_this_system()
 	update_entries jq "[.[]|select(has(\"options\"))|select(.options|match(\"root=UUID=$root_uuid\"))]"
 }
 
+find_conf_ids_by_kernel_version() {
+    local kernel_version="${1:?}"
+    local id_prefix="$entry_token-$kernel_version"
+	update_entries_for_this_system
+    jq -r --arg id_prefix "$id_prefix" '.[] | select(.id | startswith($id_prefix) and endswith(".conf")) | .id' < "$entryfile"
+}
+
 find_conf_file() {
 	local kernel_version="${1:?}"
 	local snapshot="$2"
@@ -400,6 +409,56 @@ remove_kernel()
 	update_predictions=1
 }
 
+remove_oldest_kernel()
+{
+	local snapshot="$1"
+	local root_snapshot="$2"
+	find_kernels "${snapshot}"
+	local snapshot_kernels=("${!found_kernels[@]}")
+	find_kernels "${root_snapshot}"
+	declare -A kernels=()
+	declare -a bootable_kernels
+	local id conf_files
+	for kernel in "${snapshot_kernels[@]}" "${!found_kernels[@]}"; do
+		kernels["$kernel"]=1
+	done
+
+	update_entries_for_this_system
+	while IFS= read -r id; do
+		if [[ ! $id == "$entry_token"* || ! $id == *.conf ]]; then
+			continue
+		fi
+		cleaned_id=${id#"$entry_token"-}
+		cleaned_id=${cleaned_id%.conf}
+		cleaned_id=${cleaned_id%-+([0-9]|-*[0-9])}
+
+		if [[ -n "$snapshot" ]]; then
+			cleaned_id=${cleaned_id%-*}
+		fi
+
+		bootable_kernels+=("$cleaned_id")
+	done < <(jq -r '.[].id' "$entryfile" | sort -Vu)
+
+	local oldest_removable_kernel=""
+	for kernel in "${bootable_kernels[@]}"; do
+		if [[ ! ${kernels[$kernel]+_} ]]; then
+			oldest_removable_kernel="$kernel"
+			break
+		fi
+	done
+
+	if [[ -z "$oldest_removable_kernel" ]]; then
+		log_info "Not enough disk space & no kernels to remove."
+		return 1
+	fi
+
+	conf_files=$(find_conf_ids_by_kernel_version "$oldest_removable_kernel")
+	while IFS= read -r id; do
+		bootctl unlink "$id"
+		log_info "Unlinked .conf file for ID: $id"
+	done <<< "$conf_files"
+}
+
 install_with_rollback()
 {
 	local src="${1:?}"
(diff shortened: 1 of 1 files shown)


TobiPeterG:
If no kernel can be removed, the install command fails

lnussel (changes requested):
Two remarks

If we do this I think we should make two passes over the to-be installed files instead. Ie first calculate the space they'd need. Then call a cleanup function, then try installing again if there's enough space.

The mechanism needs to be based on entry files rather than kernels though. We know the list of entries and it is sorted IIRC so we can purge the oldest by means of bootctl remove. bootctl takes care of reference counting and removing all related files, including eg device trees. We should not delete kernels or entries ourselves.

sdbootutil
```diff
@@ -269,6 +317,26 @@ install_with_rollback()
 		mv "$dst" "$dst.bak" || return "$?"
 	fi
 	rollback+=("$dst")
+
+	src_size_kb=$(du -k "$src" | cut -f1)
```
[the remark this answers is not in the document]
TobiPeterG: done :)

sdbootutil
```diff
@@ -269,6 +317,26 @@ install_with_rollback()
...
+	mkdir -p "$(dirname "$dst")" 2>/dev/null || :
```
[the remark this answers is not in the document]
TobiPeterG: done :)

sdbootutil
```diff
@@ -269,6 +317,26 @@ install_with_rollback()
...
+	if [ -d "$(dirname "$dst")" ]; then
+		available_space_kb=$(df -k "$(dirname "$dst")" | awk 'NR==2 {print $4}')
+	else
+		log_info "Directory doesn't exist: $(dirname "$dst")"
+		available_space_kb=0
+	fi
+
+	while [[ "$available_space_kb" -lt "$src_size_kb" ]]; do
```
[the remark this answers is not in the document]
TobiPeterG: done

sdbootutil
```diff
@@ -269,6 +317,26 @@ install_with_rollback()
...
+		log_info "Not enough space available in $(dirname "$dst")"
+		remove_oldest_kernel "$snapshot" "$root_snapshot" || return "$?"
+		if [ ! -d "$(dirname "$dst")" ]; then
+			mkdir -p "$(dirname "$dst")"  2>/dev/null || :
+		fi
+		if [ -d "$(dirname "$dst")" ]; then
+			available_space_kb=$(df -k "$(dirname "$dst")" | awk 'NR==2 {print $4}')
```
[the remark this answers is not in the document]
TobiPeterG: done, however, this needs explicit conversion to kb as findmnt returns size in bytes

TobiPeterG:
> Two remarks
> 
> If we do this I think we should make two passes over the to-be installed files instead. Ie first calculate the space they'd need. Then call a cleanup function, then try installing again if there's enough space.
> 
> The mechanism needs to be based on entry files rather than kernels though. We know the list of entries and it is sorted IIRC so we can purge the oldest by means of bootctl remove. bootctl takes care of reference counting and removing all related files, including eg device trees. We should not delete kernels or entries ourselves.

I changed the logic to unlinki the entries instead of manually removing files. :)

sdbootutil
```diff
@@ -307,6 +353,27 @@ install_with_rollback()
 		mv "$dst" "$dst.bak" || return "$?"
 	fi
 	rollback+=("$dst")
+
+	src_size_kb=$(du -k "$src" | { read size _; echo "$size"; })
```
lnussel, line 357: `read -r src_size  _p < <(du -k "$src")`

sdbootutil
```diff
@@ -307,6 +353,27 @@ install_with_rollback()
...
+	fi
+
+	while [ "$available_space_kb" -lt "$src_size_kb" ]; do
+		log_info "Not enough space available in ${dst%/*}"
+		remove_oldest_kernel "$snapshot" "$root_snapshot" || return "$?"
+		if [ ! -d "${dst%/*}" ]; then
+			mkdir -p "${dst%/*}"  2>/dev/null || :
+		fi
+		if [ -d "${dst%/*}" ]; then
+			available_space_kb=$(df -k "${dst%/*}" | awk 'NR==2 {print $4}')
```
lnussel, line 374: read

sdbootutil
```diff
@@ -294,10 +294,56 @@ remove_kernel()
...
+	local snapshot_kernels=("${!found_kernels[@]}")
+	find_kernels "$root_snapshot"
+	declare -A kernels=()
+	local id
+	for kernel in "${snapshot_kernels[@]}" "${!found_kernels[@]}"; do
+		kernels["$kernel"]=1
+	done
+
+
+	local bootable_kernels=($(ls "$efi_kernel_dir" | sort -V))
```
lnussel, line 313: you can just get the entries from bootctl. They are sorted so oldest should be last?

lnussel (changes requested):
As I said, two passes would be better. Ie don't hook into install_with_rollback

Also needs rebase now

sdbootutil
TobiPeterG: done

sdbootutil
TobiPeterG: done

sdbootutil
TobiPeterG: done

TobiPeterG:
So, I think I made the requested changes. Please test it, as I'm currently not home and don't have a testing environment.

I noticed that e.g. in the install_kernel the version is set to snapshot@kernel_version, even though $snapshot might not be set, I guess that would only leave @kernel_version, is that wanted?

TobiPeterG:
> So, I think I made the requested changes. Please test it, as I'm currently not home and don't have a testing environment.
> 
> I noticed that e.g. in the install_kernel the version is set to snapshot@kernel_version, even though $snapshot might not be set, I guess that would only leave @kernel_version, is that wanted?

Could test it today, should work now :)

TobiPeterG:
@lnussel did you have a chance to review it yet? :)

lnussel:
it's still trying to free space several times. IMO a better way would be to either

- calculate the required space beforehand and free if needed. We then even know how much actually was freed to check whether install fits without trying.
- or alternatively free space when install failed but then only do it once and retry once. ie not on every initrd copy

TobiPeterG:
> it's still trying to free space several times. IMO a better way would be to either
> 
> * calculate the required space beforehand and free if needed. We then even know how much actually was freed to check whether install fits without trying.
> * or alternatively free space when install failed but then only do it once and retry once. ie not on every initrd copy

I tried to rework it, I hope this is now what you are looking for :)
Every file that needs to be installed is added to an array and it's saved how much storage will be required for all of these files. We then try to make space for them, then we try to install them.
I also reworked the installation process of files a bit, I hope I didn't miss any important conditional. :)

aplanas:
@TobiPeterG I think that this feature is really important and I want have it in sdbootutil.  I was reviewing the code but it confuses me.

Sadly @lnussel will not review this PR, as he left the company to work on different projects. But if I understand him correctly the solution proposes is much more simple:

1) Around [here](https://github.com/openSUSE/sdbootutil/blob/main/sdbootutil#L614) in `install_kernel` the kernel and all the initrds are in place. The kernel in `$src` and the initrds in `$tmpdir/initrd-*`.  Note that the multiple initrd is not really working, depends on a project, mkmoduleinitrd, that is still not working.  But lets assume that there are multiple initrd in place for now

2) In this moment we know the expected size. We can add the kernel and the initrds size, and compare with the free space in the ESP.

3) If there is no free space, we need to list all the boot entries. For that we call `update_entries` to generate the entry file JSON file, and we iterate it to get the `id` (you are doing this already in your code)

4) With this list, in reverse order (I think that the filter to ID and reversing the list can be done in a single `jq`) we do `run_command_output bootctl unlink "$id"` for all the entries, except three: the last working snapshot, the current and the new one. See this: https://github.com/openSUSE/sdbootutil/blob/main/sdbootutil#L1405

In that regard `install_kernel` should not refactored, and there is no need to any global variable or keeping list of files to install. Just a some code in the line 614 (maybe a call to a function) to the sum of the kernel and initrds file and the bootctl unlink until there is free space or the list if IDs is exhausted.  This list can be generated using the complementary of the function from line 1405.

Do you want to try? Should I try and you do the review?

TobiPeterG:
> @TobiPeterG I think that this feature is really important and I want have it in sdbootutil.  I was reviewing the code but it confuses me.
> 
> Sadly @lnussel will not review this PR, as he left the company to work on different projects. But if I understand him correctly the solution proposes is much more simple:
> 
> 1) Around [here](https://github.com/openSUSE/sdbootutil/blob/main/sdbootutil#L614) in `install_kernel` the kernel and all the initrds are in place. The kernel in `$src` and the initrds in `$tmpdir/initrd-*`.  Note that the multiple initrd is not really working, depends on a project, mkmoduleinitrd, that is still not working.  But lets assume that there are multiple initrd in place for now
> 
> 2) In this moment we know the expected size. We can add the kernel and the initrds size, and compare with the free space in the ESP.
> 
> 3) If there is no free space, we need to list all the boot entries. For that we call `update_entries` to generate the entry file JSON file, and we iterate it to get the `id` (you are doing this already in your code)
> 
> 4) With this list, in reverse order (I think that the filter to ID and reversing the list can be done in a single `jq`) we do `run_command_output bootctl unlink "$id"` for all the entries, except three: the last working snapshot, the current and the new one. See this: https://github.com/openSUSE/sdbootutil/blob/main/sdbootutil#L1405
> 
> In that regard `install_kernel` should not refactored, and there is no need to any global variable or keeping list of files to install. Just a some code in the line 614 (maybe a call to a function) to the sum of the kernel and initrds file and the bootctl unlink until there is free space or the list if IDs is exhausted.  This list can be generated using the complementary of the function from line 1405.
> 
> Do you want to try? Should I try and you do the review? 


Hey, thank you very much for your message. :)
I would be happy to write the code, however, my studies, job and preparation for the openSUSEcon are currently keeping me busy (I'm unfortunately not paid for my work for SUSE ( yet ;) ), so I can't tell when I'll have the time (though I hope soon™).
If you want to write the code, I'd also be happy to review it (after all, in the end, what matters is that the feature works) :)

TobiPeterG:
done in #99

PR #73, opened 2024-03-20
Revert "Remove GRUB2 package conflict"

lnussel:
This reverts commit eaf8c97427224440fd0aea90789857ad05e38dbb.

We need to get rid of sdbootutil-rpm-scriptlets to avoid it taking prefernce over suse-module-tools-scriptlets


lnussel:
I'd to this for now to get the staging done and then try to fix this differently @aplanas

aplanas (approved):

PR #75, opened 2024-03-22
Switch from bash to rust

TobiPeterG:
This PR aims to switch from the bash script to a rust based approach. Currently, all functions available in sdbootutil are callable and have their designed parameters. A help page exists and is auto-generated.
However, the functions still lack the specific implementation, there currently is no testing framework in place and no github actions are used to e.g. build the source, test for warnings and execute tests.

Contributions are welcome :)

EDIT: Github Actions are added, will automatically build and run the specified tests

Basic tests are added

--------------------------------------------------------------

Implemented functions

- [x] help
- [ ] kernels
- [ ] snapshots
- [ ] entries
- [x] bootloader
- [ ] add-kernel
- [ ] all-all-kernels
- [ ] mkinitrd
- [ ] remove-kernel
- [ ] remove-all-kernels
- [ ] list-kernels
- [ ] list-entries
- [ ] list-snapshots
- [ ] set-default-snaphsot
- [ ] is-bootable
- [x] is-installed
- [x] install
- [x] needs-update
- [x] update
- [x] force-update
- [ ] update-predictions

Implemented functionality

- [x] Basic UI
- [ ] UI Main menu
- [ ] UI Kernels menu
- [ ] UI Entries menu
- [ ] UI Snapshots menu
- [ ] UI install/Update menu
- [x] README with contribution instructions
- [x] command line parsing


TobiPeterG:
To use, install rust (e.g. rustup and run "rustup default stable") and run "cargo run" in the project directory. This will compile the source.

lnussel:
Interesting exercise for sure :) However, before rewriting sdbootutil in something else we need to reduce it's amount of code. For example the code that installs shim etc needs to be implemented in bootctl directly.
Adding kernels and initrd is done by kernel-install upstream. It needs some brainpower to come up with a way to integrate our use case with snapshots. Lennart proposed some instance handle in the pr linked here https://progress.opensuse.org/issues/127151

aplanas:
> However, before rewriting sdbootutil in something else we need to reduce it's amount of code. For example the code that installs shim etc needs to be implemented in bootctl directly.

If you are not against the reformulation of sdbootutil in Rust, I would not condition it until what and how some features will be moved into the right components on systemd.  Both tasks has their own timeline and this PR is starting now.

TobiPeterG:
> Interesting exercise for sure :) However, before rewriting sdbootutil in something else we need to reduce it's amount of code. For example the code that installs shim etc needs to be implemented in bootctl directly.
> Adding kernels and initrd is done by kernel-install upstream. It needs some brainpower to come up with a way to integrate our use case with snapshots. Lennart proposed some instance handle in the pr linked here https://progress.opensuse.org/issues/127151

Ah I see :)
However, personally, I'm not really a fan of waiting for something that's out of control of oneself. Sdbootutil seems to become more and more important (Aeon wants to switch to systemd-boot and yast supports it already), so it would be nice to have a framework that detects regressions etc. And when the upstream changes happen, we can also reduce the code in a rust based version. However, this is all just my opinion.

So would you say I should stop working on this pr as it doesn't have a chance anyway or should I continue working and when the changes upstream happen, we just reduce code here. :)

lnussel:
you can of course dedicate your time to whatever is fun to you. I started this tool with the idea that it will become obsolete though. Eg. the fancy gui was a nice gimmick and helpful to me in the beginning, not sure it's that useful anymore. To name another example when I started kernel-install was a shell script, now it's in C so that opens new possibilities that wait to be explored.
There is a lot of value in breaking down sdbootutil's features in smaller chunks like you did. I just wouldn't start with right away reimplementing that but rather review what's really needed in a tool of ours or instead needs to go elsewhere/upstream.

TobiPeterG:
> you can of course dedicate your time to whatever is fun to you. I started this tool with the idea that it will become obsolete though. Eg. the fancy gui was a nice gimmick and helpful to me in the beginning, not sure it's that useful anymore. To name another example when I started kernel-install was a shell script, now it's in C so that opens new possibilities that wait to be explored.
> There is a lot of value in breaking down sdbootutil's features in smaller chunks like you did. I just wouldn't start with right away reimplementing that but rather review what's really needed in a tool of ours or instead needs to go elsewhere/upstream.

I see your point :)
However, I don't know if the components we deem to go upstream will go upstream anytime soon or at all.
Evaluating this, however, is of course useful. I could, for example, focus on the most important features we definitely need to get an MVP as soon as possible.
I agree, the UI is optional, however, I also see a use case for it. And as I'm trying to keep UI and logic apart, the overhead should also be minimal (and with cursive UI is quite straightforward).

I just wouldn't wait for changes to happen upstream, as we have a need for something better now and not in the future. Also, for me it looks like sdbootutil is currently still growing, not shrinking. And everything that's implemented in bash will need to be ported anyway, either upstream or to something else here.

PR #82, opened 2024-04-10
fallback for machine-id

lnussel:
In the instsys /etc/machine-id of the target is actually empty. So we need a fallback to os-release after all.


sdbootutil
```diff
@@ -357,8 +356,12 @@ settle_entry_token()
 			entry_token="$os_release_IMAGE_ID"
 			[ -n "$entry_token" ] || err "Missing IMAGE_ID"
 			;;
-		*) entry_token="$arg_entry_token" ;;
+		literal:*)
```
aplanas, line 359: is this a new option? should be documented in the help message?

lnussel: just ported from bootctl for compatibility

PR #90, opened 2024-05-09
Add new options to sdbootutil

aplanas:
The idea is to have a way to detect the default boot entry without reading the `LoaderEntryDefault` EFI var (`--default-entry`) and select the current `initrd` (`--fields=initrd`). This can be used in other scripts (`--no-colors`) that need to find those attributes

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -23,6 +23,16 @@ have_snapshots=
 # for x in vmlinuz image vmlinux linux bzImage uImage Image zImage; do
 image=
 
+color_red=
+color_end=
+if [ "$SYSTEMD_COLORS" != "false" ] && [ "$SYSTEMD_COLORS" != "0" ]; then
+	color_red="\e[31m"
+	color_green="\e[32m"
+	color_yellow="\e[33m"
+	color_bu="\e[1;4m" # bold underscore
+	color_end="\e[m"
+fi
+
 # State file for transactional systems
 state_file="/var/lib/misc/transactional-update.state"
 
@@ -103,12 +113,16 @@ helpandquit()
 			   is potentially bootable
 
 		install    Install systemd-boot and shim into ESP
+
 		needs-update
 			   Check whether the bootloader in ESP needs updating
+
 		update
 			    Update the bootloader if it's old
+
 		force-update
 			    Update the bootloader in any case
+
 		update-predictions
 			    Update TPM2 predictions
 
@@ -117,6 +131,8 @@ helpandquit()
 		snapshots  Open snapshots menu
 		entries    Open entry menu
 
+		Vairables:
+		SYSTEMD_COLORS		Set 0 to disable colored output
 	EOF
 	exit 0
 }
@@ -700,16 +716,17 @@ list_entries()
 	while read -r isdefault isreported type id root conf title; do
 		color=
 		if [ "$isdefault" = "true" ]; then
-			color="\e[1;4m"
+			color="$color_bu"
 		fi
 		if [ "$isreported" = "false" ]; then
-			color="$color\e[32m"
+			color="$color${color_green}"
 		fi
 		if [ "$type" = "loader" ]; then
-			color="$color\e[33m"
+			color="$color${color_yellow}"
 		fi
+
 		local errors=()
-		if [ -n "$verbose" ] && [ -n "$conf" ] && [  -e "$conf" ]; then
+		if [ -n "$verbose" ] && [ -n "$conf" ] && [ -e "$conf" ]; then
 			local k
 			local v
 			while read -r k v; do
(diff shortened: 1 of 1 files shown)


lnussel:
what's your use case? You could just call bootctl list yourself...

aplanas:
Doing `bootctl list --json=short` and filter the default entry and get the fields with `jq` is definitively an option, but I would not do that in the weak-modules2 from suse-module-tools. The two options that I considered was:

1) make this a feature of sdbootutil (this one), as wm2 will use it in any case
2) make wm2 to read the EFI variable to fetch the default boot entry, and read the file to separate the k/v to filter the fields (initrd in this case)

lnussel:
as agreed offline, we will introduce some `sdbootutil show-entry` command that receives a kernel version (optional snapshot number) to show the entry for a given kernel or selected fields of that entry.

aplanas:
@lnussel I added the `show-entry` command

sdbootutil
```diff
@@ -729,12 +743,42 @@ list_entries()
...
+			*) continue ;;
+		esac
+
+		if [ -z "$arg_fields" ] || [[ ${arg_fields[@]} =~ "all" ]] || [[ ${arg_fields[@]} =~ "$k" ]]; then
+			[ -z "$fields" ] || fields+="\n"
+			fields+="$k\t$v"
+		fi
+	done < "$conf"
+	[ -z "$verbose" ] || echo -e "ID\t$id"
+	echo -e "$fields"
```
lnussel, line 779: Use an array instead?

sdbootutil
```diff
@@ -729,12 +743,42 @@ list_entries()
 			done < "$conf"
 		fi
 		if [ -n "$errors" ]; then
-			echo -e "  \e[31m${errors[*]}\e[m" >&2
+			[ -n "$arg_no_colors" ] || echo -e "  \e[31m${errors[*]}\e[m" >&2
+			[ -z "$arg_no_colors" ] || echo -e "  ${errors[*]}" >&2
```
lnussel, line 747: would look better if we used something like color_red="\e[31m" and leave it empty if no colors are requested

sdbootutil
```diff
@@ -1632,9 +1679,11 @@ while true ; do
 		--arch) arg_arch="$2"; shift 2 ;;
 		--entry-token) arg_entry_token="$2"; shift 2 ;;
 		--image) image="$2"; shift 2 ;;
+		--fields) IFS=',' read -r -a arg_fields <<<"$2"; shift 2 ;;
 		--no-variables) arg_no_variables=1; shift ;;
 		--no-reuse-initrd) arg_no_reuse_initrd=1; shift ;;
 		--no-random-seed) arg_no_random_seed=1; shift ;;
+		--no-colors) arg_no_colors=1; shift ;;
```
lnussel, line 1686: systemd uses $SYSTEMD_COLORS instead of a separate command line switch. Maybe just use that too?

lnussel (changes requested):

sdbootutil
```diff
@@ -60,6 +62,8 @@ helpandquit()
 		  --arch		Manually set architecture
 		  --entry-token		Override entry token
 		  --image		Specify Linux kernel file name
+		  --fields		Comma separated list of fields
```
lnussel, line 65: https://uapi-group.org/specifications/specs/boot_loader_specification/ calls them keys. `--entry-keys`, `--show-entry-keys`?

lnussel (changes requested):

sdbootutil
aplanas: I do not think so, the ANSI color are two elements: a prefix and a suffix (`\e[m`).  I would need to introduce a new var (`color_end`) that can conditionally be empty. IMO it is clear this way

sdbootutil
aplanas: done

sdbootutil
aplanas: done

sdbootutil
aplanas: done

sdbootutil
lnussel: exactly, a marker for end also. that avoid duplicating the actual command

sdbootutil
```diff
@@ -117,6 +127,9 @@ helpandquit()
 		snapshots  Open snapshots menu
 		entries    Open entry menu
 
+		Vairables:
+		SDB_ADD_INITIAL_CMDLINE		Measure initial cmdline and initrd
+		SDB_NO_COLORS			Disable colored output
```
lnussel, line 132: I meant literall SYSTEMD_COLORS. SDB_ADD_INITIAL_CMDLINE is rather internal right, wouldn't even document that

sdbootutil
```diff
@@ -729,12 +744,42 @@ list_entries()
...
+		case "$k" in
+			title|version|sort-key|options|linux|initrd) ;;
+			*) continue ;;
+		esac
+
+		if [ -z "$arg_entry_keys" ] || [[ ${arg_entry_keys[@]} =~ "all" ]] || [[ ${arg_entry_keys[@]} =~ "$k" ]]; then
+			output+=("$k\t$v")
+		fi
+	done < "$conf"
+	printf "%b\n" "${output[@]}"
```
lnussel, line 780: Sorry for not realizing this earlier but this can be turned into a simple echo in the loop. The result isn't used anymore afterwards

lnussel (changes requested):

sdbootutil
aplanas: done

sdbootutil
aplanas: done

lnussel (approved):

PR #91, opened 2024-05-13
Add byte order mark to boot.csv

lnussel:
shim skips over it. Add it nevertheless so we can make sure it's in the correct encoding.

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -1211,7 +1211,7 @@ install_bootloader()
 		install -p -D "$bootloader" "$boot_root/EFI/BOOT/BOOT${firmware_arch^^}.EFI"
 	fi
 	# this is for shim to create the entry if missing
-	echo "${entry##*/},openSUSE Boot Manager" | iconv -f ascii -t ucs2 > "$boot_root/$boot_dst/boot.csv"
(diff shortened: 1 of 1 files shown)


sdbootutil
```diff
@@ -1211,7 +1211,7 @@ install_bootloader()
 		install -p -D "$bootloader" "$boot_root/EFI/BOOT/BOOT${firmware_arch^^}.EFI"
 	fi
 	# this is for shim to create the entry if missing
-	echo "${entry##*/},openSUSE Boot Manager" | iconv -f ascii -t ucs2 > "$boot_root/$boot_dst/boot.csv"
+	echo "${entry##*/},openSUSE Boot Manager" | { echo -ne "\xff\xfe"; iconv -f ascii -t ucs-2le; } > "$boot_root/$boot_dst/boot.csv"
```
aplanas, line 1214: Is any component checking this?

lnussel: not yet, will do so in the patch I submit to systemd to support installing from a generic directory. I wanted to require the byte order mark to avoid parsing a file that is garbage.

aplanas (approved):

PR #92, opened 2024-05-15
Call regenerate-initrd-posttrans

aplanas:
In MicroOS the initrd needs to be generated at the end of the transaction.  This commit call the regenerate script from suse-module-tools from the set_default_snapshot.

The renegerate script detect if the transaction is ongoing, and will avoid the call to sdbootutil, but if the transaction is done (for example, when called from the snapper hook) it will correctly generate a new initrd.

Note that regenerate-initrd-posttrans will move the file mark to /dev/shm if it is in a transactional system, because /run does not escape the transaction.

### DIFF
--- a/10-sdbootutil.snapper
+++ b/10-sdbootutil.snapper
@@ -46,6 +46,10 @@ set_default_snapshot()
 	if is_transactional; then
 		/usr/bin/sdbootutil add-all-kernels "$num" || :
 		/usr/bin/sdbootutil update "$num" || :
+
+		if [ -e /usr/lib/module-init-tools/regenerate-initrd-posttrans ]; then
+			/usr/lib/module-init-tools/regenerate-initrd-posttrans
+		fi
 	fi
 
 	/usr/bin/sdbootutil set-default-snapshot "$num"
--- a/10-sdbootutil.tukit
+++ b/10-sdbootutil.tukit
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+shopt -s extglob nullglob
+set -e
+
+# check whether it's a transactional system
+is_transactional()
+{
+    [ "$(stat -f -c %T /etc)" = "overlayfs" ]
+}
+
+callExt_post()
+{
+    path="$1"
+
+    if [ -e "$path"/run/regenerate-initrd ]; then
+	cp -a "$path"/run/regenerate-initrd /run
+    fi
+}
+
+h()
+{
+    echo "Available commands:"
+    echo "${!commands[@]}"
+}
+
+declare -A commands
+
+commands['callExt-post']=callExt_post
+commands['help']=h
+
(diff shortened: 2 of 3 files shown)


aplanas:
Depends on https://github.com/openSUSE/suse-module-tools/pull/103

10-sdbootutil.snapper
```diff
@@ -46,6 +46,16 @@ set_default_snapshot()
 	if is_transactional; then
 		/usr/bin/sdbootutil add-all-kernels "$num" || :
 		/usr/bin/sdbootutil update "$num" || :
+
+		# regenerate-initrd-posttrans will move the mark to
+		# /dev/shm
+		if [ -e /usr/lib/module-init-tools/regenerate-initrd-posttrans ] && \
+		   [ -e /dev/shm/transactional-update/run ]; then
```
lnussel, line 53: what's this? /dev/shm is not made for stuff like that. Might be an easy hack right now but certainly not the right thing. If there is non yet, then some handover location needs to be created. Maybe even the same /run/host hierarchy as used by systemd-nspawn

aplanas: The issue that this try to address is that in MicroOS, after a zypper dup, the regenerate-initrd-posttrans mark the creation of a new initrd in a file in /run. This file disappear once the tukit call is over (but the transaction is still open), so I need to rescue this flag from inside the same tukit call.  For that I made a change in the regenerate-initrd-posttrans script.

The same /dev/shm is used in combustion to transfer information from inside the transaction to outside, so this is not the first instance where share memory is used for this task.

The correct solution, IMO, is that t-u / tukit gain a new feature of plugins like the one from snapper, so I can inspect what is inside the transaction from the same chroot.  This still requires a communication channel from inside the chroot to outside, and I think /dev/shm is still a good candidate.

aplanas: Here is the counterpart: https://github.com/openSUSE/suse-module-tools/pull/103/files#diff-a5683575be1701b9948a3b07c0824f986dd168dc68e64e18110d21fc5a4606c2R34-R45

aplanas: Also I am implementing this tukit plugin feature (I implemented it in t-u but later I realized that tmpfs is unmounted from the chroot for each tukit call, so was not working)

lnussel: I know `/dev/shm` looks like an easy way. Doesn't make it better. combustion using it is another indication that a proper handover interface is missing. I think the topic deserves some brainstorming.
Opened https://github.com/openSUSE/transactional-update/issues/121

aplanas: Should be resolved

aplanas:
Also depends on: https://github.com/openSUSE/transactional-update/pull/122

PR #101, opened 2024-06-19
Make 'invert' a generic case

aplanas:


sdbootutil
```diff
@@ -334,7 +334,7 @@ update_entries_for_snapshot()
 update_entries_for_snapshot_invert()
 {
 	local n="$1"
-	update_entries_for_subvol "${subvol_prefix}/.snapshots/$n/snapshot" "invert"
+	update_entries_for_subvol "${subvol_prefix}/.snapshots/$n/snapshot" "not"
 }
```
TobiPeterG, line 338: This still keeps the `update_entries_for_snapshot_invert` function. Couldn't it be now merged with the `update_entries_for_snapshot` funtion?

TobiPeterG: and provide the "not" as optional parameter to it?

aplanas: I really do not want to merge it. For me is more clear to read `update_entries_for_snapshot_invert "$snapshot"`, instead of `update_entries_for_snapshot "$snapshot" "not"`. Here `not` is a jq filter and I do not want to use it directly.

PR #102, opened 2024-06-22
rollback bootloader after setting default snapshot

### LINKED ISSUE #77: Recover old bootloader and generate new predictions when doing a rollback
https://bugzilla.suse.com/show_bug.cgi?id=1222183

TobiPeterG:
basic attempt to fix #77

### DIFF
--- a/10-sdbootutil.snapper
+++ b/10-sdbootutil.snapper
@@ -45,7 +45,6 @@ set_default_snapshot()
 
 	if is_transactional; then
 		/usr/bin/sdbootutil add-all-kernels "$num" || :
-		/usr/bin/sdbootutil update "$num" || :
 
 		if [ -e /usr/lib/module-init-tools/regenerate-initrd-posttrans ]; then
 			/usr/lib/module-init-tools/regenerate-initrd-posttrans
@@ -53,6 +52,7 @@ set_default_snapshot()
 	fi
 
 	/usr/bin/sdbootutil set-default-snapshot "$num" || :
+	/usr/bin/sdbootutil update --sync "$num"
 }
 
 h()
--- a/sdbootutil
+++ b/sdbootutil
@@ -20,6 +20,7 @@ arg_no_variables=
 arg_no_reuse_initrd=
 arg_no_random_seed=
 arg_portable=
+arg_sync=
 arg_only_default=
 arg_default_snapshot=
 arg_ask_pin_or_pw=
@@ -87,6 +88,7 @@ helpandquit()
 		  --entry-keys		Comma separated list of keys
 		  --no-variables	Do not update UEFI variables
 		  --no-reuse-initrd	Always regenerate initrd
+		  --sync		Do not only update, but also downgrade
 		  --portable		Handle bootloader on portable devices
 		  --only-default	Only list the default entry
 		  --default-snapshot	[SNAPSHOT] refers to the default snapshot
@@ -151,7 +153,8 @@ helpandquit()
 			   Check whether the bootloader in ESP needs updating
 
 		update
-			    Update the bootloader if it's old
+			    Update the bootloader if it's older than the deployed version.
+				Use the --sync option to replace the deployed version also when it's newer
 
 		force-update
 			    Update the bootloader in any case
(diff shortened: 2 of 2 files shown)


10-sdbootutil.snapper
```diff
@@ -45,10 +45,10 @@ set_default_snapshot()
 
 	if is_transactional; then
 		/usr/bin/sdbootutil add-all-kernels "$num" || :
```
TobiPeterG, line 47: Why is this required? Doesn't sdbootutil automatically create the entries when with set-default-snapshot if not configs exist?

aplanas: Can be yes. But the explicit call to add-all-kernels is more informative. In any case the call here seems to be more correct. Making the call in 'set-default-spanshot' if there is not config file seems more like a optional decision.

sdbootutil
```diff
@@ -125,9 +125,15 @@ helpandquit()
 		needs-update
 			   Check whether the bootloader in ESP needs updating
 
+		out-of-sync
+			   Check whether the bootloader in ESP has a different version than on the system
+
 		update
 			    Update the bootloader if it's old
 
+		switch
```
aplanas, line 134: Would it make sense to move this to 'update', with an extra parameter like '--force' or '--sync')

sdbootutil
```diff
@@ -125,9 +125,15 @@ helpandquit()
 		needs-update
 			   Check whether the bootloader in ESP needs updating
 
+		out-of-sync
```
aplanas, line 128: I do not like the name, and I think that can be added to 'needs-update', with some '--verbose' option, that shows the current version in the ESP and the version in the appointed snapshot

aplanas (commented):
I like the approach, but IMHO one (switch) or both commands can be integrated in the existing ones

sdbootutil
TobiPeterG: done

sdbootutil
TobiPeterG: done

TobiPeterG:
This makes needs_update way more verbose, personally, I like that.

TobiPeterG:
ping
Can I rebase for merge?

aplanas:
> Can I rebase for merge?

Rebase it, sure. But I will need more time to review :disappointed:

TobiPeterG:
Hey, I finally found the time to rebase, feel free to test and review :)

sdbootutil
```diff
@@ -20,6 +20,7 @@ arg_no_variables=
 arg_no_reuse_initrd=
 arg_no_random_seed=
 arg_portable=
+arg_sync=
```
aplanas, line 23: Is there a situation when we do not want this to be the default?

sdbootutil
```diff
@@ -1543,6 +1554,17 @@ install_bootloader()
 	update_predictions=1
 }
 
+bootloader_update()
+{
+	local status=0
+	bootloader_needs_update "${1:-$root_snapshot}" || status=$?
+	if [[ $status -eq 0 ]]; then
```
aplanas, line 1561: `[ "$status" = "0" ]` ? I think this is the idiom used in this codebase

sdbootutil
```diff
@@ -1543,6 +1554,17 @@ install_bootloader()
...
+		install_bootloader "${1:-$root_snapshot}"
+	elif [ -n "$arg_sync" ] && [[ $status -eq 2 ]]; then
```
aplanas, line 1563: `[ "$status" = "2" ]` ?

sdbootutil
```diff
@@ -1458,10 +1461,18 @@ bootloader_needs_update()
 	nv="$(bootloader_version "$(find_bootloader "$snapshot")")"
 	[ -n "$v" ] || return 1
 	log_info "system version $nv"
-	systemd-analyze compare-versions "$v" lt "$nv" 2>/dev/null || return 1
+	systemd-analyze compare-versions "$v" "$nv" 2>/dev/null
+	local status="$?"
 	bldr_name=$(bootloader_name "$snapshot")
-	log_info "$bldr_name needs to be updated"
-	return 0
+	if [ "$status" == "11" ]; then
```
aplanas, line 1467: single '=' here and in the nexts?

aplanas (approved):

TobiPeterG:
Oh thank you, I'm sorry that I didn't have a look at your review yet, I'm sick for quite a while now. :(

I planned on working on sdbootutil, but I was laying in bed instead. I will be though when I'm back on full power :)

aplanas:
Please, take care of yourself!!

PR #106, opened 2024-07-01
Support portable installation of bootloader

TobiPeterG:
This is useful to create portable drives, so the bootloader entry isn't created permenantly.

@sysrich needs this feature for his TIK installer.

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -21,6 +21,7 @@ arg_no_variables=
 arg_no_reuse_initrd=
 arg_no_random_seed=
 arg_ask_pin=
+arg_portable=
 have_snapshots=
 # for x in vmlinuz image vmlinux linux bzImage uImage Image zImage; do
 image=
@@ -76,6 +77,7 @@ helpandquit()
 		  --no-variables	Do not update UEFI variables
 		  --no-reuse-initrd	Always regenerate initrd
 		  --ask-pin		Ask recovery PIN for re-enrollment
+		  --portable		Handle bootloader on portable devices
 		  -v, --verbose		More verbose output
 		  -h, --help		This screen
 
@@ -1342,7 +1344,7 @@ install_bootloader()
 		log_info "Installing $bldr_name with shim into $boot_root"
 		entry="$boot_dst/shim.efi"
 		for i in MokManager shim; do
-			install -p -D "$prefix$shimdir/$i.efi" "$boot_root$boot_dst/$i.efi"
+			[ -n "$arg_portable" ] || install -p -D "$prefix$shimdir/$i.efi" "$boot_root$boot_dst/$i.efi"
 		done
 		install -p -D "$bootloader" "$boot_root$boot_dst/grub.efi"
 
@@ -1354,14 +1356,15 @@ install_bootloader()
 	else
 		log_info "Installing $bldr_name into $boot_root"
 		entry="$boot_dst/${bootloader##*/}"
-		install -p -D "$bootloader" "$boot_root$entry"
+		[ -n "$arg_portable" ] || install -p -D "$bootloader" "$boot_root$entry"
 		install -p -D "$bootloader" "$boot_root/EFI/BOOT/BOOT${firmware_arch^^}.EFI"
 	fi
+
 	# this is for shim to create the entry if missing
-	echo "${entry##*/},openSUSE Boot Manager" | { echo -ne "\xff\xfe"; iconv -f ascii -t ucs-2le; } > "$boot_root/$boot_dst/boot.csv"
+	[ -n "$arg_portable" ] || echo "${entry##*/},openSUSE Boot Manager" | { echo -ne "\xff\xfe"; iconv -f ascii -t ucs-2le; } > "$boot_root$boot_dst/boot.csv"
 
 	mkdir -p "$boot_root/$entry_token"
-	echo "$entry_token" > "$boot_root/$boot_dst/installed_by_sdbootutil"
+	echo "$entry_token" > "$boot_root$boot_dst/installed_by_sdbootutil"
 	mkdir -p "/etc/kernel"
 	[ -s /etc/kernel/entry-token ] || echo "$entry_token" > /etc/kernel/entry-token
 	update_random_seed
@@ -1395,7 +1398,7 @@ install_bootloader()
 	fi
 
 	# Create boot menu entry if it does not exist
-	[ -n "$arg_no_variables" ] || efibootmgr | grep -q 'Boot.*openSUSE Boot Manager' || efibootmgr -q --create --disk "$drive" --part "$partno" --label "openSUSE Boot Manager" --loader "$entry" || true
+	[ -n "$arg_no_variables" ] || [ -n "$arg_portable" ] || efibootmgr | grep -q 'Boot.*openSUSE Boot Manager' || efibootmgr -q --create --disk "$drive" --part "$partno" --label "openSUSE Boot Manager" --loader "$entry" || true
 
 	# This action will require to update the PCR predictions
 	update_predictions=1
@@ -1827,10 +1830,12 @@ generate_tpm2_predictions_pcrlock()
 
 	# 630-shim-efi-application is not part of the pcrlock standards
 	# TODO: move to shim-pcrlock.rpm
+	local entry="shim.efi"
+	[ -f "$boot_dst/BOOT${firmware_arch^^}.EFI" ] || entry="BOOT${firmware_arch^^}.EFI"
 	pcrlock \
 	    lock-pe \
 	    --pcrlock=/var/lib/pcrlock.d/630-shim-efi-application.pcrlock.d/generated.pcrlock \
-	    "${boot_root}${boot_dst}/shim.efi"
+	    "${boot_root}${boot_dst}/${entry}"
 
 	# 640-boot-loader-efi-application is not part of the pcrlock
 	# standards
@@ -1990,7 +1995,7 @@ main_menu()
 
 ####### main #######
 
-getopttmp=$(getopt -o hc:v --long help,flicker,verbose,esp-path:,entry-token:,arch:,image:,entry-keys:,no-variables,no-reuse-initrd,no-random-seed,ask-pin,all -n "${0##*/}" -- "$@")
+getopttmp=$(getopt -o hc:v --long help,flicker,verbose,esp-path:,entry-token:,arch:,image:,entry-keys:,no-variables,no-reuse-initrd,no-random-seed,ask-pin,portable,all -n "${0##*/}" -- "$@")
 eval set -- "$getopttmp"
 
 while true ; do
@@ -2007,6 +2012,7 @@ while true ; do
 		--no-reuse-initrd) arg_no_reuse_initrd=1; shift ;;
 		--no-random-seed) arg_no_random_seed=1; shift ;;
 		--ask-pin) arg_ask_pin=1; shift ;;
+		--portable) arg_portable=1; shift ;;
 		--all) arg_all_entries=1; shift ;;
                 --) shift ; break ;;
                 *) echo "Internal error!" ; exit 1 ;;
@@ -2063,7 +2069,13 @@ case "$firmware_arch" in
 esac
 
 # XXX: Unify both in /EFI/opensuse?
-if is_sdboot; then
+if [ -n "$arg_portable" ]; then
+	if [ ! -d "$boot_root/EFI/systemd" ] && [ ! -d "$boot_root/EFI/opensuse" ]; then
+		boot_dst="/EFI/BOOT"
+	else
+		err "Bootloader is already installed permanently"
+	fi
+elif is_sdboot; then
 	boot_dst="/EFI/systemd"
 elif is_grub2; then
 	boot_dst="/EFI/opensuse"


sdbootutil
```diff
@@ -1323,6 +1325,7 @@ install_bootloader()
 		prefix="/.snapshots/${root_snapshot}/snapshot"
 	fi
 	local bootloader bldr_name blkpart drive partno
+	local efi_dst="/EFI/BOOT"
```
aplanas, line 1328: you defined boot_dst to have this value, should be used instead of redefining it here?

sdbootutil
```diff
@@ -1338,27 +1341,26 @@ install_bootloader()
 	drive="/dev/${drive##*/}"
 	read -r partno < "/sys/class/block/${blkpart##*/}"/partition
 
-	if [ -e "$prefix$shimdir/shim.efi" ]; then
+	if [ -e "$prefix/$shimdir/shim.efi" ]; then
```
aplanas, line 1344: are you sure? `$shimdir` should already start with `/`

sdbootutil
```diff
@@ -1338,27 +1341,26 @@ install_bootloader()
...
 		log_info "Installing $bldr_name with shim into $boot_root"
 		entry="$boot_dst/shim.efi"
-		for i in MokManager shim; do
-			install -p -D "$prefix$shimdir/$i.efi" "$boot_root$boot_dst/$i.efi"
-		done
-		install -p -D "$bootloader" "$boot_root$boot_dst/grub.efi"
+		install -p -D "$prefix/$shimdir/MokManager.efi" "$boot_root/$boot_dst/MokManager.efi"
+		[ -n "$arg_portable" ] || install -p -D "$prefix/$shimdir/shim.efi" "$boot_root/$boot_dst/shim.efi"
```
aplanas, line 1348: I would prefer something more generic, if `shim` is missing do not install it, but do not make dependent of it is portable or not

sdbootutil
```diff
@@ -1338,27 +1341,26 @@ install_bootloader()
...
+		install -p -D "$bootloader" "$boot_root/$boot_dst/grub.efi"
```
aplanas, line 1349: I think that you are adding '/' where they do not belong?

aplanas:
> This is useful to create portable drives, so the bootloader entry isn't created permenantly.

You mean that they are not registered in the EFI boot entries?

sdbootutil
```diff
@@ -1338,27 +1341,26 @@ install_bootloader()
...
 
 		# boot entry point
-		for i in MokManager fallback; do
-			install -p -D "$prefix$shimdir/$i.efi" "$boot_root/EFI/BOOT/$i.efi"
-		done
-		install -p -D "$prefix$shimdir/shim.efi" "$boot_root/EFI/BOOT/BOOT${firmware_arch^^}.EFI"
+		[ -n "$arg_portable" ] || install -p -D "$prefix/$shimdir/MokManager.efi" "$boot_root/$efi_dst/MokManager.efi"
```
aplanas, line 1352: Should be more generic, and if MokManager is not installed in the image, just do not use it.

Also please drop the extra `/`

sdbootutil
```diff
@@ -1827,10 +1829,12 @@ generate_tpm2_predictions_pcrlock()
 
 	# 630-shim-efi-application is not part of the pcrlock standards
 	# TODO: move to shim-pcrlock.rpm
+	local shim_name="shim.efi"
+	[ -z "$arg_portable" ] || shim_name="BOOT${firmware_arch^^}.EFI"
```
aplanas, line 1833: This is very wrong, I think. Is the to measure the shim.efi and only the shim.efi, not the BOOT<arch>.EFI

sdbootutil
```diff
@@ -2063,7 +2068,9 @@ case "$firmware_arch" in
 esac
 
 # XXX: Unify both in /EFI/opensuse?
-if is_sdboot; then
+if [ -n "$arg_portable" ]; then
+	boot_dst="/EFI/BOOT"
```
aplanas, line 2072: I do not know how a protable environment is created, but do exist /EFI/systemd or /EFI/opensuse in this case?

aplanas (changes requested):

sdbootutil
TobiPeterG: When it's not a portable install, the installation process needs to write to this location anyway. And since it's accessed on multiple occasions, I exported the path to this variable

sdbootutil
TobiPeterG: Yes, I just noticed that in some places the slash is already there.
A double slash doesn't break the parsing afaik, but makes it look more consistent imo

sdbootutil
TobiPeterG: Shim gets installed further down when it's a portable install, but the filename changes.

sdbootutil
TobiPeterG: See above
Technically unnecessary, but it also doesn't break stuff and at least for me makes it more clear that it's still a path

sdbootutil
TobiPeterG: I just don't install it here as it got installed in the exact same place above already when it's a portable installation

sdbootutil
TobiPeterG: The boot.efi is the shim.efi in a portable installation, just renamed for uefi to find it. :)

sdbootutil
TobiPeterG: No, these locations don't exist, as the portable installation mode doesn't require them. Only the default /EFI/BOOT should be populated afaik

sdbootutil
aplanas: Do not change this then, will help me for the review : )

sdbootutil
aplanas: Do not change it then

sdbootutil
TobiPeterG: Because if boot_dst is something different than /efi/boot efi_dst still has to be /EFI/boot

sdbootutil
aplanas: My point is that if mokutil is installed, then should be copied, and if not, then the boot entry should not be registered. This make a portable system something that is a consequence of what you have installed too

For example, when building an image with KIWI, some actions should not be done but there is no need to pass an extra parameter to do that.  In that regard, I am not sure why this is really needed, because the KIWI image should be an instance of a portable system in some regard

sdbootutil
aplanas: Good point, yes. What if shim is not installed?

sdbootutil
aplanas: In that case there is no need to pass a parameter, as can be inferred, can't it?

sdbootutil
aplanas: Understood, it is a constant. Then do not use a variable for that. It is more clear to see that this is a constant when it is used

sdbootutil
TobiPeterG: How?
When installing systemd-boot for the first time, the /EFI/systemd-boot folder never exist, in both the portable and persistent case

sdbootutil
TobiPeterG: How exactly would this work?
Mokmanager is always required when secure boot is wanted, isn't it?

TobiPeterG: > My point is that if mokutil is installed, then should be copied, and if not, then the boot entry should not be registered. This make a portable system something that is a consequence of what you have installed too
> 
> For example, when building an image with KIWI, some actions should not be done but there is no need to pass an extra parameter to do that.  In that regard, I am not sure why this is really needed, because the KIWI image should be an instance of a portable system in some regard


So on systems without shim, mokmanager will also not be available afaik, even though we might want to install systemd-boot permanently.

sdbootutil
aplanas: If shim is installed then BOOT<arch>.EFI is shim, and grub2.efi is the bootloader. If not then BOOT<arch>.EFI is the bootloader

If this is the case then the predictions should be adjusted, and the boot loader installation logic should be adjusted too. Or I am missing something?

sdbootutil
aplanas: This is true. Would make sense to validate it in a running system? I men, if the parameter is passed but /EFI/systemd or /EFI/opensuse is present, then did something wrong happen?

TobiPeterG: Good question.
I passed a test mount, will test it in a VM later. Though I don't expect issues as I hope that boot_dst is used to determine the bootloader location everywhere.

TobiPeterG: > This is true. Would make sense to validate it in a running system? I men, if the parameter is passed but /EFI/systemd or /EFI/opensuse is present, then did something wrong happen?


Oh I just noticed: why would one test this. When the bootloader is installed as portable, why would there already be a /EFI/systemd folder?

aplanas: I was thinking in the reverse case. A normal one, but the user uses `--portable`, that will change the wrong place

TobiPeterG: Ah yes, this would change the wrong place, but I guess calling install without the --portable flag would fix it. I mean that's why it's portable, it doesn't do anything persistent.

aplanas: I am confused, changing /boot/efi/EFI/BOOT in the ESP of a running system is pretty persistent IMO ...

TobiPeterG: Yes, this part is persistent, but not in the UEFI

The worst thing I can imagine that could happen is that there are unnecessary files in /EFI/boot, but nothing that would prevent the system from booting.
But I'll test

TobiPeterG: > I am confused, changing /boot/efi/EFI/BOOT in the ESP of a running system is pretty persistent IMO ...

I tested it, what I did:
sdbootutil install
reboot
sdbootutil --portable install
reboot
sdbootutil install
reboot

The system is always bootable, after sdbootutil --portable install there are some extra files in /efi/boot, but they don't interfere with the boot process

sdbootutil
TobiPeterG: > If shim is installed then BOOT.EFI is shim, and grub2.efi is the bootloader. If not then BOOT.EFI is the bootloader
> 
> If this is the case then the predictions should be adjusted, and the boot loader installation logic should be adjusted too. Or I am missing something?

Should be fixed now, the shim.efi file is only measured when not in portable mode

TobiPeterG:
```
pcrlock \
	    lock-pe \
	    --pcrlock=/var/lib/pcrlock.d/640-boot-loader-efi-application.pcrlock.d/generated.pcrlock \
	    "${boot_root}${boot_dst}/grub.efi"
```

This is generally wrong when shim is not used, then the bootloader will not have this name.
But this should be fixed in another PR IMO

sdbootutil
aplanas: Please. Do not use this variable. It is confusing.

It is more clear `"$boot_root/EFI/BOOT/BOOT${firmware_arch^^}.EFI"` `"$boot_root/$efi_dst/BOOT${firmware_arch^^}.EFI"`. For one the reader can see part of the path, and for other the variable makes to believe that it can change.

sdbootutil
aplanas: Please, drop the extra `/`.  Do not include this change for this PR, we can discuss it in a different one.

sdbootutil
aplanas: I still believe that this is better:

```
for i in MokManager shim; do
  install -p -D "$prefix$shimdir/$i.efi" "$boot_root$boot_dst/$i.efi" || true
done
```

or

```
for i in MokManager shim; do
  [ ! -f "$prefix$shimdir/$i.efi" ] || install -p -D "$prefix$shimdir/$i.efi" "$boot_root$boot_dst/$i.efi"
done
```

This cover your case and the case when the image does not have shim not mok manager

sdbootutil
aplanas: For this PR do not include the `/`, lets focus only in the main feature.

sdbootutil
aplanas: same than before

sdbootutil
```diff
@@ -1338,27 +1341,26 @@ install_bootloader()
...
 		entry="$boot_dst/${bootloader##*/}"
-		install -p -D "$bootloader" "$boot_root$entry"
-		install -p -D "$bootloader" "$boot_root/EFI/BOOT/BOOT${firmware_arch^^}.EFI"
+		[ -n "$arg_portable" ] || install -p -D "$bootloader" "$boot_root$entry"
+		install -p -D "$bootloader" "$boot_root/$efi_dst/BOOT${firmware_arch^^}.EFI"
 	fi
+
 	# this is for shim to create the entry if missing
-	echo "${entry##*/},openSUSE Boot Manager" | { echo -ne "\xff\xfe"; iconv -f ascii -t ucs-2le; } > "$boot_root/$boot_dst/boot.csv"
+	[ -n "$arg_portable" ] || echo "${entry##*/},openSUSE Boot Manager" | { echo -ne "\xff\xfe"; iconv -f ascii -t ucs-2le; } > "$boot_root/$boot_dst/boot.csv"
```
aplanas, line 1363: having this file does not break any portable system, right?

sdbootutil
```diff
@@ -1389,13 +1391,13 @@ install_bootloader()
...
-		mkdir -p "$boot_root$boot_dst/$(uname -m)-efi"
-		cp -a "$prefix$grub2moddir/bli.mod" "$boot_root$boot_dst/$(uname -m)-efi"
+		[ -e "$boot_root/$efi_dst/grub.cfg" ] || cp "$boot_root/$boot_dst/grub.cfg" "$boot_root/$efi_dst/grub.cfg"
+		mkdir -p "$boot_root/$boot_dst/$(uname -m)-efi"
+		cp -a "$prefix/$grub2moddir/bli.mod" "$boot_root/$boot_dst/$(uname -m)-efi"
 	fi
 
 	# Create boot menu entry if it does not exist
-	[ -n "$arg_no_variables" ] || efibootmgr | grep -q 'Boot.*openSUSE Boot Manager' || efibootmgr -q --create --disk "$drive" --part "$partno" --label "openSUSE Boot Manager" --loader "$entry" || true
+	[ -n "$arg_no_variables" ] || [ -n "$arg_portable" ] || efibootmgr | grep -q 'Boot.*openSUSE Boot Manager' || efibootmgr -q --create --disk "$drive" --part "$partno" --label "openSUSE Boot Manager" --loader "$entry" || true
```
aplanas, line 1400: IMO this is the main change, can be?

sdbootutil
```diff
@@ -1827,7 +1829,7 @@ generate_tpm2_predictions_pcrlock()
 
 	# 630-shim-efi-application is not part of the pcrlock standards
 	# TODO: move to shim-pcrlock.rpm
-	pcrlock \
+	[ -n "$arg_portable" ] || pcrlock \
```
aplanas, line 1832: the condition should be that `[ ! -f "${boot_root}${boot_dst}/shim.efi" ] || ` but I guess that this is not enough.

The system will measure "${boot_root}${boot_dst}/BOOT<some_arch>.EFI` so this  should be measured and not ignored

sdbootutil
aplanas: In that case check that if `--portable` is passed, then no systemd nor opensuse is present, and error if this is the case.

It is not OK to alter /boot/efi/EFI/BOOT in non portable systems. I would argue that a more correct approach is to consider this directory when the bootloader is updated, but that will be a matter for a different PR

aplanas (changes requested):

aplanas:
> This is generally wrong when shim is not used, then the bootloader will not have this name. But this should be fixed in another PR IMO

Yes, but until now it is save to guess that shim is used. But you are right, this needs to be changed for a more general case.

sdbootutil
TobiPeterG: done

sdbootutil
TobiPeterG: done

sdbootutil
TobiPeterG: I don't think so. This will create a shim.efi in the /EFI/BOOT directory, even though we don't need one.

sdbootutil
TobiPeterG: done

sdbootutil
TobiPeterG: AFAIK it does.
This csv tells the UEFI to create the entry when it doesn't exist already.

sdbootutil
TobiPeterG: Yes and no. This command would create the UEFI boot entry for the current system, so on reboot it is visible. The csv AFAIK is used when the entry doesn't exist yet for the UEFI to auto-create it.

sdbootutil
TobiPeterG: The BOOTx64.efi is the shim file in the /EFI/BOOT directory when installed as portable, there is no shim.efi to measure. We don't need it. It's the BOOTx64.efi
When shim is not available, BOOTx64.efi is the bootloader. In this case, we also don't need it. In this case, however, there is no grub.efi to be measured, so the next pcrlock command is wrong.

sdbootutil
```diff
@@ -1341,27 +1343,26 @@ install_bootloader()
...
 		log_info "Installing $bldr_name into $boot_root"
 		entry="$boot_dst/${bootloader##*/}"
-		install -p -D "$bootloader" "$boot_root$entry"
+		[ -n "$arg_portable" ] || install -p -D "$bootloader" "$boot_root$entry"
 		install -p -D "$bootloader" "$boot_root/EFI/BOOT/BOOT${firmware_arch^^}.EFI"
 	fi
+
 	# this is for shim to create the entry if missing
-	echo "${entry##*/},openSUSE Boot Manager" | { echo -ne "\xff\xfe"; iconv -f ascii -t ucs-2le; } > "$boot_root/$boot_dst/boot.csv"
+	[ -n "$arg_portable" ] || echo "${entry##*/},openSUSE Boot Manager" | { echo -ne "\xff\xfe"; iconv -f ascii -t ucs-2le; } > "$boot_root$boot_dst/boot.csv"
```
aplanas, line 1362: Right, seems that this file is processed by fallback.efi and added into the NVRAM

sdbootutil
aplanas: My point is that it is not OK to break the prediction, so this needs more work.

Your argument that because BOOT<arch>.EFI can be shim is wrong. Of course we need it. Measured boot will load BOOT<arch>.EFI and create an event log for it in PCR#4. So we need to create a component for it via `lock-pe`. 

The change is that the EFI binary will be under a different name.

For now lets assume that shim is installed. In this case the most simple change should be something like this:

```
	# 630-shim-efi-application is not part of the pcrlock standards
	# TODO: move to shim-pcrlock.rpm
        local entry="$boot_dst/shim.efi"
        [ -f "$boot_dst/BOOT${firmware_arch^^}.EFI" ] || entry="$boot_dst/BOOT${firmware_arch^^}.EFI"
	pcrlock \
	    lock-pe \
	    --pcrlock=/var/lib/pcrlock.d/630-shim-efi-application.pcrlock.d/generated.pcrlock \
	    "${boot_root}${entry}"
```

But this is not entirely correct, because the shim package maybe was not installed. Do you get the idea?

aplanas (changes requested):
Thanks a lot for simplifying the changes, this helps me a lot.

sdbootutil
```diff
@@ -1341,27 +1343,26 @@ install_bootloader()
 	if [ -e "$prefix$shimdir/shim.efi" ]; then
 		log_info "Installing $bldr_name with shim into $boot_root"
 		entry="$boot_dst/shim.efi"
-		for i in MokManager shim; do
-			install -p -D "$prefix$shimdir/$i.efi" "$boot_root$boot_dst/$i.efi"
-		done
+		install -p -D "$prefix/$shimdir/MokManager.efi" "$boot_root$boot_dst/MokManager.efi"
```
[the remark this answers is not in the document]
TobiPeterG: I refactored it, is it better now?

sdbootutil
```diff
@@ -1341,27 +1343,26 @@ install_bootloader()
...
+		[ -n "$arg_portable" ] || install -p -D "$prefix$shimdir/shim.efi" "$boot_root$boot_dst/shim.efi"
 		install -p -D "$bootloader" "$boot_root$boot_dst/grub.efi"
 
 		# boot entry point
-		for i in MokManager fallback; do
-			install -p -D "$prefix$shimdir/$i.efi" "$boot_root/EFI/BOOT/$i.efi"
-		done
+		[ -n "$arg_portable" ] || install -p -D "$prefix$shimdir/MokManager.efi" "$boot_root/EFI/BOOT/MokManager.efi"
+		install -p -D "$prefix$shimdir/fallback.efi" "$boot_root/EFI/BOOT/fallback.efi"
```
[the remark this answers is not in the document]
TobiPeterG: done I hope :)

sdbootutil
TobiPeterG: should be done now :)

sdbootutil
TobiPeterG: done :)
Should probably be refactored in another PR

sdbootutil
```diff
@@ -1342,26 +1344,27 @@ install_bootloader()
 		log_info "Installing $bldr_name with shim into $boot_root"
 		entry="$boot_dst/shim.efi"
 		for i in MokManager shim; do
-			install -p -D "$prefix$shimdir/$i.efi" "$boot_root$boot_dst/$i.efi"
+  			[ -n "$arg_portable" ] || install -p -D "$prefix$shimdir/$i.efi" "$boot_root$boot_dst/$i.efi"
```
aplanas, line 1347: nit: initial space

sdbootutil
```diff
@@ -1342,26 +1344,27 @@ install_bootloader()
...
 		done
 		install -p -D "$bootloader" "$boot_root$boot_dst/grub.efi"
 
 		# boot entry point
 		for i in MokManager fallback; do
-			install -p -D "$prefix$shimdir/$i.efi" "$boot_root/EFI/BOOT/$i.efi"
+  			install -p -D "$prefix$shimdir/$i.efi" "$boot_root/EFI/BOOT/$i.efi"
```
aplanas, line 1353: nit: initial space

sdbootutil
```diff
@@ -1827,10 +1830,12 @@ generate_tpm2_predictions_pcrlock()
 
 	# 630-shim-efi-application is not part of the pcrlock standards
 	# TODO: move to shim-pcrlock.rpm
+	local entry="shim.efi"
+    [ -f "$boot_dst/BOOT${firmware_arch^^}.EFI" ] || entry="BOOT${firmware_arch^^}.EFI"
```
aplanas, line 1834: nit: indent

aplanas (approved):
LGTM, if you can fix the nits lets merge it and fix the measurements in the next PR

TobiPeterG:
> LGTM, if you can fix the nits lets merge it and fix the measurements in the next PR

LGTM :)

aplanas (approved):

aplanas:
@TobiPeterG thanks!

PR #110, opened 2024-07-12
Turn off colors when the shell it not interactive

danyspin97:
Allow overide by using SYSTEMD_COLORS.

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -28,7 +28,12 @@ image=
 
 color_red=
 color_end=
-if [ "$SYSTEMD_COLORS" != "false" ] && [ "$SYSTEMD_COLORS" != "0" ]; then
+# The shell is interactive and the colors haven't been disabled forcefully
+# or the shell is not interactive but the colors have been forcefully enabled
+if { [[ "$-" == *i* ]] \
+    && [ "$SYSTEMD_COLORS" != "false" ] && [ "$SYSTEMD_COLORS" != "0" ]; } \
+    || [[ "$-" != *i* ]] \
+    && { [ "$SYSTEMD_COLORS" = "true" ] || [ "$SYSTEMD_COLORS" = "1" ]; }; then
 	color_red="\e[31m"
 	color_green="\e[32m"
 	color_yellow="\e[33m"


sdbootutil
```diff
@@ -28,7 +28,12 @@ image=
 
 color_red=
 color_end=
-if [ "$SYSTEMD_COLORS" != "false" ] && [ "$SYSTEMD_COLORS" != "0" ]; then
+# The shell is interactive and the colors haven't been disabled forcefully
+# or the shell is not interactive but the colors have been forcefully enabled
+if { [[ $- == *i* ]] \
+    && [ "$SYSTEMD_COLORS" != "false" ] && [ "$SYSTEMD_COLORS" != "0" ]; } \
+    || [[ $- != *i* ]] \
+    && { [ "$SYSTEMD_COLORS" == "true" ] || [ "$SYSTEMD_COLORS" == "1" ]; }; then
```
aplanas, line 36: LGTM but the style so far is use `=` instead of `==` , and maybe `[[ "$-" =~ "i" ]]`

danyspin97: `[[ "$-" =~ "i" ]]` triggered a warning with shellcheck and it suggested to use `=` instead.

sdbootutil
```diff
@@ -28,7 +28,12 @@ image=
 
 color_red=
 color_end=
-if [ "$SYSTEMD_COLORS" != "false" ] && [ "$SYSTEMD_COLORS" != "0" ]; then
+# The shell is interactive and the colors haven't been disabled forcefully
+# or the shell is not interactive but the colors have been forcefully enabled
+if { [[ "$-" = *i* ]] \
+    && [ "$SYSTEMD_COLORS" != "false" ] && [ "$SYSTEMD_COLORS" != "0" ]; } \
+    || [[ $- != *i* ]] \
```
aplanas, line 35: Missing quotes here?

sdbootutil
aplanas: Very interesting, do you have the ID?

aplanas: oh never mind, I found it

aplanas: If I read correctly the suggestion is to use `[[ "$-" =~ i ]]` ?

danyspin97: Updated the PR with the correct version.

sdbootutil
danyspin97: Yea, they are not really needed but they don't hurt either.

sdbootutil
```diff
@@ -28,7 +28,12 @@ image=
 
 color_red=
 color_end=
-if [ "$SYSTEMD_COLORS" != "false" ] && [ "$SYSTEMD_COLORS" != "0" ]; then
+# The shell is interactive and the colors haven't been disabled forcefully
+# or the shell is not interactive but the colors have been forcefully enabled
+if { [[ "$-" =~ i ]] \
```
aplanas, line 33: Ah Danilo, very sorry to waste your time.

After reading from you that shellcheck complained I changed the sdbootutil =~ operators to match the expectations. I did it here:

https://github.com/openSUSE/sdbootutil/pull/111/commits/01a56be188ef7bf570ddece106d5bbadfce9aa39

So following your initial change, I moved to `[[ "$-" == *"i"* ]]` as recommended in https://www.shellcheck.net/wiki/SC2076

danyspin97: I see, I will update the commit as well then.

aplanas (approved):

PR #111, opened 2024-07-12
Implement basic enroll / unenroll commands

aplanas:


aplanas:
cc @TobiPeterG I think that you are interested in this feature
cc @danyspin97

TobiPeterG:
Hey,
Sorry that I didn't respond yet. I'll have a look at it tomorrow. :)

aplanas:
I merged it a bit early, as it is fixing a bug that is hitting the devel project

sdbootutil
```diff
@@ -1988,14 +2021,404 @@ generate_tpm2_predictions()
...
+		if [[ "$opts" != *"$option"* ]]; then
+			[ -z "$opts" ] && opts="$option" || opts="$opts,$option"
+			# crypttab has changed so initrd needs to be updated
+			update_initrd=1
+		fi
+		echo "$name $device ${key:-none} $opts" >> "$crypttab"
+	done < /etc/crypttab
+
+	mv "$crypttab" /etc/crypttab
+	chmod 644 /etc/crypttab
```
TobiPeterG, line 2085: This enrolls the TPM to all devices listed in crypttab. I have multiple encrypted drives and I only want to enroll the root partition. Maybe we should check if cr_root exists in the crypttab and use only that.

TobiPeterG: Or alternatively, try to find what device the root fs is placed on. 
For example
This
`findmnt -n -o SOURCE /`

Has the output
`/dev/mapper/cr_root[/@/.snapshots/475/snapshot]`

sdbootutil
```diff
@@ -1988,14 +2021,404 @@ generate_tpm2_predictions()
...
+		"password")
+			enroll_password "$dev" "$pin_or_pw"
+			;;
+
+		*)
+			err "Unexpected parameter for --method=: $arg_method"
+			;;
+	esac
+}
+
```
TobiPeterG, line 2242: Make a bit clearer what methods are available and/or set a default method (e.g. tpm2)

sdbootutil
```diff
@@ -1988,14 +2021,404 @@ generate_tpm2_predictions()
...
+enroll()
+{
+	[ -e /etc/crypttab ] || { log_info "/etc/crypttab not found. No encrypted devices?"; return 0; }
+	[ -e /usr/bin/systemd-cryptenroll ] || { log_info "systemd-cryptenroll not found"; return 0; }
+	detect_luks2 || { log_info "No LUKS2 devices found"; return 0; }
```
TobiPeterG, line 2247: Why does this return 0? Shouldn't it be treated as error?

sdbootutil
```diff
@@ -1988,14 +2021,404 @@ generate_tpm2_predictions()
...
+	if [ "$arg_method" = "tpm2" ] || [ "$arg_method" = "tpm2+pin" ]; then
+		if [ -z "${FDE_SEAL_PCR_LIST}" ]; then
+			if is_sdboot; then
+				FDE_SEAL_PCR_LIST="0,2,4,7,9"
+			elif is_grub2; then
+				FDE_SEAL_PCR_LIST="0,2,4,7,8,9"
+			else
+				err "Bootloader not detected"
+			fi
+		fi
```
TobiPeterG, line 2284: I might be stupid, when is FDE_SEAL_PCR_LIST set before enrollment?

sdbootutil
```diff
@@ -1988,14 +2021,404 @@ generate_tpm2_predictions()
...
+		[ -e /etc/sysconfig/fde-tools ] || echo "FDE_SEAL_PCR_LIST=${FDE_SEAL_PCR_LIST}" > /etc/sysconfig/fde-tools
+		# shellcheck disable=SC1091
+		. /etc/sysconfig/fde-tools
```
TobiPeterG, line 2287: Why is this required?

TobiPeterG (commented):
I added some thoughts :)

sdbootutil
aplanas: I share your concern here. The encrypted partitions should not be only rootfs, as in some systems /var is also a different partition and this should also be covered.

But I see why some users do not want to use the TPM2 for other partitions.

I am not sure how to separate them.  IIUC the ones in /etc/crypttab are the one that are registered as system partitions, and this would make sense to be under TPM2.  If the user want to manage their own partition, they can do that later.

If this way of thinking is wrong, then we need to separate the ones in /etc/crypttab, filtering the ones that have some attribute

sdbootutil
aplanas: yes

sdbootutil
aplanas: not really? I guess that LUKS2 is not the only encryption format in linux?

sdbootutil
TobiPeterG: This makes sense, but AFAIK the password would be shared between different disks, so they would all unlock, as long as they have the same password.
Or are there systems where we want to unlock different drives that all also have a different password?

sdbootutil
TobiPeterG: > not really? I guess that LUKS2 is not the only encryption format in linux?

Yes, but don't we only support LUKS2 encrypted drives? So I would expect an error when I call the enroll command with a non-LUKS2 drive.

sdbootutil
aplanas: The password should be shared, yes, at least that was my intention

TobiPeterG: Alright, so all encrypted drives have the same passowrd? Then it would be enough to only enroll the root disk AFAIK :)

sdbootutil
aplanas: if crypttab is missing there is no error, just a notification, if cryptenroll is missing also a notification without error.

Error should indicate that some operation fails, but if some of those elements no operation really failed as the enrollment was not done. If the operation can be done but fails, then this should be an error

For example, imagine that the installer calls sdbootutil enroll always at the end but crypttab is not populated because the user did not selected any disk to encrypt. Then the sdbootutil enroll should be noop

TobiPeterG: Alright, makes sense :)

sdbootutil
aplanas: I am a bit lost. The idea is that when you enroll a password, all the encrypted partitions will share the password. When you enroll a TPM2, all the encrypted partitions will share the same TPM2, and the same for TPM2+PIN or FIDO2

Why this will only affect only the rootfs?

TobiPeterG: You are absolutely right, I'm just dumb lol
Yes, we would need to enroll all encrypted drives that should be unlocked.

Would it be suitable to assume that all drives that should be enrolled that with cr_? cr_root is the root partition on Tumbleweed.
Users could then name their drives anything else, but not start with cr_root (could be made clear with a comment in the crypttab file)

aplanas: The cr_XXX pattern name is interesting. The issue is that is the installer the one that set the name, but I will keep in mind. Can work.

sdbootutil
aplanas: for pcrlock contains the PCR list (FDE_SEAL_PCR_LIST)

sdbootutil
```diff
@@ -1988,14 +2021,404 @@ generate_tpm2_predictions()
...
+{
+	local dev fstype
+	[ -z "$luks2_devices" ] || return 0
+	while read -r dev fstype; do
+		[ "$fstype" = 'crypto_LUKS' ] || continue
+		cryptsetup isLuks --type luks2 "$dev" || continue
+		luks2_devices+=("$dev")
+	done < <(lsblk --noheadings -o PATH,FSTYPE)
+	have_luks2
+}
```
TobiPeterG, line 2047: Oh I just noticed that changing the crypttab file isn't enough, this will still detect all encrypted drives, no matter what.
Can't we change this to only check drives that are in the crypttab and not enroll to ALL encrypted drives of a system?

sdbootutil
aplanas: yeah it is confusing ... This is like: first set the default values, if the fde-tools file is missing, then use those default values, but if it is present, use the ones in the file

sdbootutil
TobiPeterG, line 2085: Shouldn't this be 600? I guess it's pretty dangerous to allow everyone to read the file.

sdbootutil
aplanas: the check if is set or not early is because you can set it before calling the script

sdbootutil
TobiPeterG: Yes, but isn't the file sourced when creating the predictions anyway?

sdbootutil
aplanas: You mean iterate over the devices that are only in crypttab to decide if there is a LUKS2 device in the system?

sdbootutil
aplanas: enroll is an operation, and prediction is a different operation

when doing the enroll the pcr list can be missing, so we set default values and create the file, when doing the predictions we require the pcr list

and we can update the predictions as an independent operation

sdbootutil
TobiPeterG: Exactly :)
Otherwise all my encrypted devices get enrolled and I as a user won't be able to prevent this.

sdbootutil
TobiPeterG: I don't have a source other than this is the permission set by default for the file on Tumbleweed:
`-rw------- 1 root root 679  1. Dez 2023  /etc/crypttab`

It also makes sense IMO as one can set a keyfile or password in there and one might not want all users to know that.

sdbootutil
TobiPeterG, line 2047: This function could be replaced by this
```
detect_luks2() {
    local dev fstype name crypttab_dev device_path
    [ "${#luks2_devices[@]}" -eq 0 ] || return 0

    # Read the crypttab file and process each line
    while read -r name crypttab_dev _; do
        # Skip lines that don't start with "cr_"
        [[ "$name" =~ ^cr_ ]] || continue

        # Handle UUID format in crypttab
        if [[ "$crypttab_dev" =~ ^UUID= ]]; then
            device_path=$(blkid -U "${crypttab_dev#UUID=}")
        else
            device_path="$crypttab_dev"
        fi

        # Match devices from lsblk output with crypttab devices
        while read -r dev fstype; do
            [ "$fstype" = 'crypto_LUKS' ] || continue
            if [ "$dev" = "$device_path" ]; then
                if cryptsetup isLuks --type luks2 "$dev"; then
                    luks2_devices+=("$dev")
                fi
            fi
        done < <(lsblk --noheadings -o PATH,FSTYPE)
    done < /etc/crypttab
    
    have_luks2
}
```

It reads the crypttab file and matches only lines that start with cr_ and gets the device path. Worked in my testing.

sdbootutil
```diff
@@ -1988,14 +2021,404 @@ generate_tpm2_predictions()
...
 
 		generate_tpm2_predictions_pcrlock "${FDE_SEAL_PCR_LIST}"
 	fi
 }
 
+have_luks2()
+{
+	[ "${#luks2_devices[@]}" -gt 0 ]
+}
+
```
TobiPeterG, line 2036: I'm also not sure if we need this function. We could just check this directly instead of calling the function. IMO a function is useful when more than one command is executed.

sdbootutil
TobiPeterG: Also, AFAIK the -z $luks2_devices call always fails as it's any array and not a string, doesn't it?

TobiPeterG (commented):
I provided examples on how the functions could look, do you like this approach?
If yes, I would open a PR :)

sdbootutil
TobiPeterG: @aplanas any thoughts on this?

PR #112, opened 2024-07-15
Add --only-default for list-entries command

danyspin97:


aplanas (commented):
There is a `show-entry` command, that has `--entry-keys` to list only a subset of the elements of the entries.  This command has a VERSION parameter (for the kernel), why not to make this parameter optional and then only show the default one?

sdbootutil
```diff
@@ -864,6 +867,55 @@ list_entries()
...
+					errors+=("/.snapshot/$snapshot/snapshot does not exist")
+				fi
+			fi
+		done < "$conf"
+	fi
+	if [ -n "$errors" ]; then
+		echo -e "  ${color_red}${errors[*]}${color_end}" >&2
+	fi
+	echo -e "$color$id${verbose:+: $title}${color_end}"
+}
```
aplanas, line 917: Is a copy of list_entries, so maybe the name is confusing too. You want to list the name of the default entry but "show" is used to show the content of the entry.

If what you want is to list only the name of the default entry, then why not to add a `--only-default` parameter, and update the list_entries command?

danyspin97:
> There is a show-entry command, that has --entry-keys to list only a subset of the elements of the entries. This command has a VERSION parameter (for the kernel), why not to make this parameter optional and then only show the default one?

I didn't want to change the usage of show-entry command, but that is definitely useful.

aplanas:
> > There is a show-entry command, that has --entry-keys to list only a subset of the elements of the entries. This command has a VERSION parameter (for the kernel), why not to make this parameter optional and then only show the default one?
> 
> I didn't want to change the usage of show-entry command, but that is definitely useful.

Right, I realized in my second comment. `show_` is to show the content of an entry, and `list_` to list the name of the entries, that is what confused me.  I still think that a full command similar to list_entries can be avoided just updating list_entries.

danyspin97:
> Right, I realized in my second comment. show_ is to show the content of an entry, and list_ to list the name of the entries, that is what confused me. I still think that a full command similar to list_entries can be avoided just updating list_entries.

Showing the content of the default entry is good for me as well, it saves me from reading it manually. Maybe it is more useful in general than just showing the path to the default entry.

sdbootutil
danyspin97: I followed your advice and added a `only-default` option.

sdbootutil
```diff
@@ -2034,6 +2037,7 @@ while true ; do
 		--no-random-seed) arg_no_random_seed=1; shift ;;
 		--ask-pin) arg_ask_pin=1; shift ;;
 		--portable) arg_portable=1; shift ;;
+        --only-default) only_default=1; shift ;;
```
aplanas, line 2040: mixing tabs / spaces

sdbootutil
```diff
@@ -829,6 +830,8 @@ list_entries()
 		color=
 		if [ "$isdefault" = "true" ]; then
 			color="$color_bu"
+		elif [ "$only_default" -eq 1 ]; then
+			continue
```
aplanas, line 834: Arguments are prefixed with `arg_` and defined at the beginning of the script

`[ -z "$arg_only_default" ] || continue`

sdbootutil
aplanas: name should be arg_only_default

aplanas (changes requested):

sdbootutil
danyspin97: Fixed.

sdbootutil
danyspin97: Fixed.

sdbootutil
danyspin97: > `[ -z "$arg_only_default" ] || continue`

It is okay to leave the check as it is now? I am not sure how to add that without also checking for `$isdefault` again, or use a `else` clause instead of `elif`.

aplanas: Yeah sure, in this case the idiom used in this project seems to be `elif [ -n "$arg_only_default" ]; then`

aplanas (approved):

PR #113, opened 2024-07-18
Fix --only-default option not being recognized

danyspin97:


aplanas (commented):
Yes sorry. I fixed it in main

PR #116, opened 2024-07-29
fix portable installation

TobiPeterG:
The fallback.efi should not be added when installing in portable mode

Needed by @sysrich

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -1389,9 +1389,8 @@ install_bootloader()
 		install -p -D "$bootloader" "$boot_root$boot_dst/grub.efi"
 
 		# boot entry point
-		for i in MokManager fallback; do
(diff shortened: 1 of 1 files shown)


aplanas:
LGTM. Why fallback should not be there, and the MokManager is OK to be there?

TobiPeterG:
> LGTM. Why fallback should not be there, and the MokManager is OK to be there?

AFAIK, yes. MOKManger is used to provide Machine Owner keys, I guess it might be useful on portable installations as well.
But maybe @sysrich knows more about this.

aplanas:
but do you know why fallback.efi should not be present?

TobiPeterG:
> but do you know why fallback.efi should not be present?

@sysrich had issues with it not booting with it. The shim documentation also lists that fallback.efi should not be present:
https://github.com/rhboot/shim/blob/main/README.fallback

It also lists that MokManager should be present

aplanas:
Great documentation!

PR #119, opened 2024-08-08
Reorder PIN and PW search

aplanas:


TobiPeterG:
Why is this needed?

aplanas:
> Why is this needed?

There is this observation that I agree with (see end of the comment): https://bugzilla.suse.com/show_bug.cgi?id=1228820#c4

PR #125, opened 2024-08-27
Warn when pcrlock.json or tpm2-pcr-signature.json are missing

danyspin97:


aplanas (commented):
LGTM but I was thinking to validate the cryptenroll output.  This can be tricky as calling the enrollment for pcrlock, if fails, will trigger also the enrollment via pcr-oracle

danyspin97:
> I was thinking to validate the cryptenroll output.

I.e. checking that cryptenroll successfully enrolled the signature in the TPM?

aplanas:
> > I was thinking to validate the cryptenroll output.
> 
> I.e. checking that cryptenroll successfully enrolled the signature in the TPM?

Right. This can be done here: https://github.com/openSUSE/sdbootutil/blob/main/sdbootutil#L2178-L2205

danyspin97:
Should I create a new PR with this change?

aplanas:
As you wish. For me makes sense new commit in this PR

aplanas:
LGTM! Did you try to make it fail, so see if the pcr-oracle fallback kicks in?

danyspin97:
> LGTM! Did you try to make it fail, so see if the pcr-oracle fallback kicks in?

I haven't tried yet. To make it fail, I guess can just remove the pcrlock.json file, rigth?

aplanas:
right, you can comment the code that checks pcrlock.json and trace the code with bash -x

danyspin97:
Deleting `pcrlock.json` makes function `enroll_pcrlock` fail as expected. Then the error path is executed, I updated the function names.

`enrolling_pcroracle` though always fail with the following message:

```
Failed to read TPM PCR public key: no such file or directory
```

I tried again, this time just making the function `enroll_pcrlock` exit unsuccessfully before calling `systemd-cryptenroll`. I get the same message on a newly created MicroOS image. I am investigating right now.

aplanas:
Good catch!

danyspin97:
```bash
	NEWPIN="$tpm_pin" systemd-cryptenroll \
		--tpm2-device=auto \
		"${extra_args[@]}" \
		--tpm2-public-key=/etc/systemd/tpm2-pcr-public-key.pem \
		--tpm2-public-key-pcrs="${FDE_SEAL_PCR_LIST}" \
		"$dev"
```

I found the culprit: file `/etc/systemd/tpm2-pcr-public-key.pem` is missing.

danyspin97:
> This is generated via calling generate_rsa_key

Which is not called if `have_pcrlock` is true.

```
		if ! have_pcrlock && have_pcr_oracle; then
			generate_rsa_key
		fi
```

aplanas (approved):

aplanas:
Thanks!

PR #126, opened 2024-08-29
Add --default-snapshot parameter

aplanas:
Instead of using the [SNAPSHOT] parameter, we can now refer to the default snapshot with this new option.  It will replace the `root_snapshot` variable with the default snapshot instead the current one.


aplanas:
Fixes bsc#1224773

PR #130, opened 2024-09-05
Replace root=UUID= with root=device

aplanas:


aplanas:
@Vogtinator this will take care of switching from UUID to device name (only if encrypted)

PR #131, opened 2024-09-06
Show recovery PIN generated by systemd-pcrlock

danyspin97:


sdbootutil
```diff
@@ -1944,6 +1944,7 @@ generate_tpm2_predictions_pcrlock()
 	local pin
 	local extra=()
 	local keyid
+	local r
```
aplanas, line 1947: rename to `output`? Maybe declare it close to where it is used?

sdbootutil
```diff
@@ -1954,16 +1955,28 @@ generate_tpm2_predictions_pcrlock()
...
 		extra=("--recovery-pin=yes")
+	else
+		# No PIN was provided, systemd-pcrlock will generate one
+		# Add this argument to show it
+		extra=("--recovery-pin=show")
 	fi
 
-	PIN="$pin" pcrlock --pcr="$pcrs" "${extra[@]}" make-policy || {
+	r="$(PIN="$pin" pcrlock --pcr="$pcrs" "${extra[@]}" make-policy)"
+	if [ $? != 0 ]; then
```
aplanas, line 1965: I think you can still use the old construct?

sdbootutil
```diff
@@ -1954,16 +1955,28 @@ generate_tpm2_predictions_pcrlock()
...
 		if [ -z "$pin" ]; then
 			echo "Please, provide the recovery PIN to register the new policy"
 		else
 			echo "Provided PIN incorrect or TPM2 locked after too many retries"
 		fi
-	}
+	elif [ -z "$pin" ]; then
+		if ! echo "$r" | grep "recovery PIN"; then
+			echo "Unable to find the generated recovery PIN"
+		elif [ -x /usr/bin/qrencode ]; then
```
aplanas, line 1975: The recovery PIN should be showed even if qrencode is missing, right? Like the recovery key.

sdbootutil
```diff
@@ -1954,16 +1955,28 @@ generate_tpm2_predictions_pcrlock()
...
+				echo "You can also scan it with your mobile phone:"
+				echo "$r" | grep "recovery PIN" | cut -d":" -f2 | qrencode -t utf8i
```
aplanas, line 1977: `cut' is not idiomatic in this codebase, there are some bash way to split the string

aplanas: Only the QR is presented? Should be both I think

sdbootutil
danyspin97: Fixed.

sdbootutil
danyspin97: Fixed.

sdbootutil
danyspin97: Replaced `cut` with an idiomatic code.

aplanas (approved):

PR #132, opened 2024-09-09
Enroll a new recovery key

danyspin97:
Add a method to enroll a new recovery key.  The only things missing are reading the key from `systemd-cryptenroll --recovery-key` and it should be ready and testing the command.

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -2017,7 +2017,7 @@ generate_tpm2_predictions_pcrlock()
 	local pin
 	local extra=()
 	local keyid
-	keyid="$(keyctl id %user:cryptenroll 2> /dev/null)" || true
+	keyid="$(keyctl id %user:recoverypin 2> /dev/null)" || true
 	if [ -n "$arg_ask_pin_or_pw" ]; then
 		read -r -s -p "Recovery PIN: " pin
 		extra=("--recovery-pin=yes")
@@ -2056,9 +2056,9 @@ generate_tpm2_predictions_pcrlock()
 
 				# Add the generated recovery PIN to the kernel keyring, so that it is available to
 				# sdbootutil --method=recovery-key
-				echo -n "$pin" | keyctl padd user pin @u > /dev/null
+				echo -n "$pin" | keyctl padd user recoverykey @u > /dev/null
 				# Make the key expire after 2 minutes
-				keyctl timeout %user:pin 120
+				keyctl timeout %user:recoverykey 120
 		fi
 	fi
 
@@ -2509,7 +2509,9 @@ enroll()
 		fi
 	fi
 
-	if [ "$arg_method" == "recovery-key" ] && tmp_key_id=$(keyctl id %user:pin); then
+	# When calling --method=tpm2 and a recovery PIN is generated, it is also set in
+	# %user:recoverykey to be used as recovery key by --method=recovery-key
+	if [ "$arg_method" == "recovery-key" ] && tmp_key_id=$(keyctl id %user:recoverykey); then
 		recovery_key_id="$tmp_key_id"
 	fi
 
--- a/sdbootutil
+++ b/sdbootutil
@@ -31,7 +31,6 @@ have_snapshots=
 # for x in vmlinuz image vmlinux linux bzImage uImage Image zImage; do
 image=
 recovery_key_id=
-tpm2_policy_enrolled=
 unlock_extra_arg=
 
 color_red=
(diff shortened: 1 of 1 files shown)


danyspin97:
Added the multidisk implementation.

danyspin97:
Added the opposite flow: generated recovery PIN -> set recovery key.

sdbootutil
```diff
@@ -2257,6 +2268,50 @@ enroll_password()
...
+	local dev="$1"
+
+	echo "Enrolling with recovery key: $dev"
+
+	local recovery_key
+	# This function will be called for every device. Let systemd generate a secure recovery key
+	# only the first time and if there is no recovery pin selected (%user:pin or %user:cryptenroll).
+	# Unlock the LUKS2 header by using the TPM2 slot
+	if [ -z "$recovery_key_id" ]; then
+			if recovery_key=$(systemd-cryptenroll --wipe-slot=recovery --recovery-key --unlock-tpm2-device=auto "$dev"); then
```
aplanas, line 2282: somehow the indent is wrong here

aplanas: Can you split the `if` and the assignment? For me is not clear if the condition is over the value of `recovery_key` or the return code of `systemd-cryptenroll`

sdbootutil
```diff
@@ -2257,6 +2268,50 @@ enroll_password()
...
+			# sdbootutil --method=tpm2
+			read -r recovery_key_id < <(echo -n "$recovery_key" | keyctl padd user cryptenroll @u)
+			keyctl timeout %user:cryptenroll 120
+		else
+			echo "Unable to generate a recovery key for $dev"
+		fi
+	else
+		# A recovery key has already been generated, use it for all the devices
+		recovery_key="$(keyctl pipe "$recovery_key_id")"
+		if ! tmp_rk=$(systemd-cryptenroll --wipe-slot=recovery --recovery-key --unlock-tpm2-device=auto "$dev"); then
```
aplanas, line 2300: same

sdbootutil
```diff
@@ -2257,6 +2268,50 @@ enroll_password()
...
+			echo "Unable to access device $dev"
+			return 1
+		fi
+		# systemd-cryptenroll always generated a random recovery key, but we want $recovery_key
+		# Replace it by using cryptsetup
+		read -r -a split < <(systemd-cryptenroll /dev/vda3 | grep recovery)
+		keyslot="${split[0]}"
+		# cryptsetup can only read the new passphrase from a keyfile
+		echo -n "$recovery_key" > recovery_key
+		cryptsetup luksChangeKey --key-slot "$keyslot" "$dev" recovery_key < <(echo "$tmp_rk")
```
aplanas, line 2310: what if is the first enrollment? This key will not be present

Use tmpfs to store the recover key file. And I think that the echo can be removed via `<<<"$tmp_rk"`

aplanas (changes requested):

sdbootutil
danyspin97: Using `$?` is marked as warning by shellcheck though.

sdbootutil
danyspin97: The key is always generated above (`systemd-cryptenroll --recovery-key`) and then changed with a new one manually, so we can safely call `luksChangeKey`.

aplanas: I am missing something then. What I am reading is that if "recovery_key_id" is empty then it will generate a recovery key. This will the case for the first device.

But the second device does not have a recovery key keyslot, and at this point "recovery_key_id" is not empty.

aplanas: aggg I see it now, sorry. Why not to enroll one using only cryptsetup?

sdbootutil
aplanas: This requires that the TPM2 enrollment happen first, and there is no guarantee of that (nor that is only a FIDO2 enrollment, or a tpm2+pin)

sdbootutil
danyspin97: Because this way it's systemd-cryptenroll that uses the TPM2 to unlock the device instead of doing it manually (we don't know the current password / recovery key at this time) and the user might not know either.

aplanas: I am not sure that I agree. In the old model d-e-t generate a recovery key and use it to drive the enrollment. We should use the same approach but with this "management-key" or "enrollment-key". This key is generated during installation and used only to drive the initial enrollment, and then dropped.

This means that yes, the installed (d-e-t has this role) knows the password to use.

Depends on the TPM2 is not realistic, as TPM2 is optional. I can decide to enroll the FIDO2 key and a recovery key, or only to enroll a simple old password.

sdbootutil
danyspin97: > This requires that the TPM2 enrollment happen first, and there is no guarantee of that.

This is true, but we know that there is a TPM2 valid slot during first-enrollment. If the user knows the recovery key can first do the TPM2 enrollment and then enroll a new recovery key.

> FIDO2 enrollment

Maybe we can add a `--unlock` argument that defaults to `auto` and will use the first one of the available slots (TPM2, FIDO2, passphrase, recovery key) by checking `systemd-cryptenroll` output.

> or a tpm2+pin

In this case shouldn't systemd ask for the PIN directly? I don't see any other options specific for TPM2.

aplanas (commented):
There are some nits appointed, and other I discovered later (error messages in certain uses cases).

I will merge and fix in a subsequent commit.

aplanas (approved):

PR #134, opened 2024-09-13
Use chroot instead of --sysroot in dracut

aplanas:


aplanas:
cc @aafeijoo-suse

sdbootutil
```diff
@@ -610,6 +610,25 @@ reuse_initrd()
 	return 1
 }
 
+mount_chroot()
+{
+	local snapshot_dir="$1"
+
+	mount -t tmpfs -o size=10m tmpfs "$snapshot_dir/run"
+	for i in proc dev sys var tmp; do
```
danyspin97, line 618: Is there a specific reason to mount `/var` and `/tmp`?

aplanas: /tmp to make sure that the generated initrd can be recovered from outside, and /var for /var/tmp, that is used as a tmpdir.

Do you see as optionals?

danyspin97: I guessed it was the case for `/var` but I wasn't sure, since I was expecting `var/tmp` directly. It's fine either way.

Thanks!

aafeijoo-suse (commented):
First, thanks for working on this. TL;DR it works, but with some drawbacks that we must evaluate.

But:
- No dependency for the ESP device, because /boot/efi is not mounted in the chroot environment. For a regular use case, it's ok, because it's not needed. But, do you need this for `dracut-pcr-signature`?
- The root device is not identified as root device by dracut ([`mountpoint /` does not return 0](https://github.com/openSUSE/dracut/blob/91732c01f7ed6d8f9047f8c2ef11e8379e93e7b2/dracut.sh#L1667), so [the device under / is not added to `root_devs`](https://github.com/openSUSE/dracut/blob/91732c01f7ed6d8f9047f8c2ef11e8379e93e7b2/dracut.sh#L1671), so [no `/etc/cmdline.d/95root-dev.conf`](https://github.com/openSUSE/dracut/blob/91732c01f7ed6d8f9047f8c2ef11e8379e93e7b2/modules.d/95rootfs-block/module-setup.sh#L39-L41) is added), but it's identified as a required device, so it's added an initqueue dependency.

E.g.:

```
localhost:~ # blkid /dev/vda3
/dev/vda3: LABEL="ROOT" UUID="045fa833-1b00-47b4-b476-ddacf2e817ac" UUID_SUB="d4658f2a-6da6-4586-8baf-96ef466ec51b" BLOCK_SIZE="4096" TYPE="btrfs" PARTLABEL="p.lxroot" PARTUUID="16e6418e-6374-4cdc-bca3-59607e42428b"
```

Diff between using --sysroot and chroot:

```
localhost:~ # diff initrd-before.txt initrd-after.txt 
47d46
< etc/cmdline.d/95root-dev.conf
67,68c66,67
< etc/systemd/system/dev-disk-by\x2duuid-ED90\x2d4A05.device.d
< etc/systemd/system/dev-disk-by\x2duuid-ED90\x2d4A05.device.d/timeout.conf
---
> etc/systemd/system/dev-disk-by\x2duuid-045fa833\x2d1b00\x2d47b4\x2db476\x2dddacf2e817ac.device.d
> etc/systemd/system/dev-disk-by\x2duuid-045fa833\x2d1b00\x2d47b4\x2db476\x2dddacf2e817ac.device.d/timeout.conf
74c73
< etc/systemd/system/initrd.target.wants/dev-disk-by\x2duuid-ED90\x2d4A05.device
---
> etc/systemd/system/initrd.target.wants/dev-disk-by\x2duuid-045fa833\x2d1b00\x2d47b4\x2db476\x2dddacf2e817ac.device
1119,1120d1117
< usr/sbin/fsck.fat
< usr/sbin/fsck.vfat
1171c1168
< var/lib/dracut/hooks/emergency/80-\x2fdev\x2fdisk\x2fby-uuid\x2fED90-4A05.sh
---
> var/lib/dracut/hooks/emergency/80-\x2fdev\x2fdisk\x2fby-uuid\x2f045fa833-1b00-47b4-b476-ddacf2e817ac.sh
1174c1171
< var/lib/dracut/hooks/initqueue/finished/devexists-\x2fdev\x2fdisk\x2fby-uuid\x2fED90-4A05.sh
---
> var/lib/dracut/hooks/initqueue/finished/devexists-\x2fdev\x2fdisk\x2fby-uuid\x2f045fa833-1b00-47b4-b476-ddacf2e817ac.sh
```

What are we missing without [95root-dev.conf](https://github.com/openSUSE/dracut/blob/91732c01f7ed6d8f9047f8c2ef11e8379e93e7b2/modules.d/95rootfs-block/module-setup.sh#L43-L63)? dracut appends these options to the kernel command line, although I think some of them are duplicates.

```
localhost:~ # lsinitrd -f etc/cmdline.d/95root-dev.conf /boot/efi/opensuse-microos/6.10.8-1-default/initrd-a699b5da1139c6c31eeb36e29c29ab7a9e1f746c
 root=UUID=045fa833-1b00-47b4-b476-ddacf2e817ac rootfstype=btrfs rootflags=ro,relatime,seclabel,discard=async,space_cache=v2,subvolid=275,subvol=/@/.snapshots/8/snapshot,subvol=@/.snapshots/8/snapshot
```

aplanas:
> First, thanks for working on this. TL;DR it works, but with some drawbacks that we must evaluate.

Thanks for the review!

>     * No dependency for the ESP device, because /boot/efi is not mounted in the chroot environment. For a regular use case, it's ok, because it's not needed. But, do you need this for `dracut-pcr-signature`?

For `dracut-pcr-signature` should not be needed during the `initrd` creation. The modules and scripts that should be included in the `initrd` are in the rootfs. The json file is later included in the initrd but in memory, never on disk.

Do you recommend to mount /boot/efi in the chroot env?

>     * The root device is not identified as root device by dracut

Is this fixed by adding `'--add-device' "$root_device"` in the dracut parameters?

aafeijoo-suse:
> Do you recommend to mount /boot/efi in the chroot env?

Personally, no. I've been trying to remove this (IMO useless) dependency for quite some time now: https://github.com/dracutdevs/dracut/issues/2044

> > ```
> > * The root device is not identified as root device by dracut
> > ```
> 
> Is this fixed by adding `'--add-device' "$root_device"` in the dracut parameters?

No, that doesn't work, and it's not necessary because the initrd has a required dependency on the root device. The problem is around the `mountpoint /` code. We could think about manually installing a conf file with the [missing kernel command line options](https://github.com/openSUSE/dracut/blob/91732c01f7ed6d8f9047f8c2ef11e8379e93e7b2/modules.d/95rootfs-block/module-setup.sh#L43-L63) under `/etc/cmdline.d/` using `--include`? Although I'm not sure if all are necessary, e.g., `root=` is the same in our case, and the other options seem redundant.

```
localhost:~ # for i in $(</proc/cmdline); do echo $i | grep -e "root=" -e "rootfstype=" -e "rootflags="; done
root=UUID=045fa833-1b00-47b4-b476-ddacf2e817ac
rootflags=subvol=@/.snapshots/9/snapshot
```
```
localhost:~ # for i in $(lsinitrd -f etc/cmdline.d/95root-dev.conf /boot/efi/opensuse-microos/6.10.5-1-default/initrd-cfcd53aa0dc8bda31781c9e3d9628526c2e16257); do echo $i | grep -e "root=" -e "rootfstype=" -e "rootflags="; done
root=UUID=045fa833-1b00-47b4-b476-ddacf2e817ac
rootfstype=btrfs
rootflags=ro
```

TobiPeterG:
> > Do you recommend to mount /boot/efi in the chroot env?
> 
> Personally, no. I've been trying to remove this (IMO useless) dependency for quite some time now: https://github.com/dracutdevs/dracut/issues/2044
> 
> > > ```
> > > * The root device is not identified as root device by dracut
> > > ```
> > 
> > Is this fixed by adding `'--add-device' "$root_device"` in the dracut parameters?
> 
> No, that doesn't work, and it's not necessary because the initrd has a required dependency on the root device. The problem is around the `mountpoint /` code. We could think about manually installing a conf file with the [missing kernel command line options](https://github.com/openSUSE/dracut/blob/91732c01f7ed6d8f9047f8c2ef11e8379e93e7b2/modules.d/95rootfs-block/module-setup.sh#L43-L63) under `/etc/cmdline.d/` using `--include`? Although I'm not sure if all are necessary, e.g., `root=` is the same in our case, and the other options seem redundant.
> 
> ```
> localhost:~ # for i in $(</proc/cmdline); do echo $i | grep -e "root=" -e "rootfstype=" -e "rootflags="; done
> root=UUID=045fa833-1b00-47b4-b476-ddacf2e817ac
> rootflags=subvol=@/.snapshots/9/snapshot
> ```
> ```
> localhost:~ # for i in $(lsinitrd -f etc/cmdline.d/95root-dev.conf /boot/efi/opensuse-microos/6.10.5-1-default/initrd-cfcd53aa0dc8bda31781c9e3d9628526c2e16257); do echo $i | grep -e "root=" -e "rootfstype=" -e "rootflags="; done
> root=UUID=045fa833-1b00-47b4-b476-ddacf2e817ac
> rootfstype=btrfs
> rootflags=ro
> ```

Just a quick comment, afaik the dracut repo you linked is dead.
This is the repo Arch switched to:
https://github.com/dracut-ng/dracut-ng
So maybe have a try there, maybe they already fixed it :)

There is an issue open on the dracut repo you linked.
Afaik there are also some efforts on openSUSE's side to switch to it :)

aafeijoo-suse:
> Just a quick comment, afaik the dracut repo you linked is dead. This is the repo Arch switched to: https://github.com/dracut-ng/dracut-ng So maybe have a try there, maybe they already fixed it :)

No, it was not fixed, and after almost 2 years I gave up: https://github.com/dracut-ng/dracut-ng/pull/335

PR #147, opened 2024-10-22
Enroll only cr_ prefixed devices

TobiPeterG:
This will fix the issue that all devices in crypttab are enrolled, even though this might not be desired. This implementation also will not replace the crypttab file, but only update it. This also creates a backup of the old file.

Please verify that this works as exected on different machines and use-cases

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -2136,16 +2136,34 @@ have_luks2()
 	[ "${#luks2_devices[@]}" -gt 0 ]
 }
 
-detect_luks2()
-{
-	local dev fstype
-	[ -z "$luks2_devices" ] || return 0
-	while read -r dev fstype; do
-		[ "$fstype" = 'crypto_LUKS' ] || continue
-		cryptsetup isLuks --type luks2 "$dev" || continue
-		luks2_devices+=("$dev")
-	done < <(lsblk --noheadings -o PATH,FSTYPE)
-	have_luks2
+detect_luks2() {
+    local dev fstype name crypttab_dev device_path
+    [ "${#luks2_devices[@]}" -eq 0 ] || return 0
+
+    # Read the crypttab file and process each line
+    while read -r name crypttab_dev _; do
+        # Skip lines that don't start with "cr_"
+        [[ "$name" =~ ^cr_ ]] || continue
+
+        # Handle UUID format in crypttab
+        if [[ "$crypttab_dev" =~ ^UUID= ]]; then
+            device_path=$(blkid -U "${crypttab_dev#UUID=}")
+        else
+            device_path="$crypttab_dev"
+        fi
+
+        # Match devices from lsblk output with crypttab devices
+        while read -r dev fstype; do
+            [ "$fstype" = 'crypto_LUKS' ] || continue
+            if [ "$dev" = "$device_path" ]; then
+                if cryptsetup isLuks --type luks2 "$dev"; then
+                    luks2_devices+=("$dev")
+                fi
+            fi
+        done < <(lsblk --noheadings -o PATH,FSTYPE)
+    done < /etc/crypttab
+    
+    have_luks2
 }
 
 have_tpm2()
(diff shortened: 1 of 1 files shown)


TobiPeterG:
This doesn't rewrite the remove_crypttab_option yet, but that should, if the functionality proposed here is wanted, should also be done :)

aplanas:
I was thinking in a different approach instead of the name of the luks2 device. For example, we could use an `x-sdbootutil.` namespace attribute in /etc/crypttab to registers the ones that will be automatically tracked by sdbootutil. Or enumerate then in the /etc/sysconfig/fdetools file

The problem tracking only cr_* is that is a convention outside the expectation of the users (configuration files and such)

sdbootutil
```diff
@@ -2158,33 +2176,31 @@ have_fido2()
...
-			# crypttab has changed so initrd needs to be updated
-			arg_no_reuse_initrd=1
-		fi
-		echo "$name $device ${key:-none} $opts" >> "$crypttab"
-	done < /etc/crypttab
+    # Use sed to add the option if it's not already present and line starts with cr_
+    sed -i -E "/^cr_/ {
+        s/(\S+\s+\S+\s+\S+\s*)(.*)/\1\2/
+        /$option/! {
+            s/$/,${option}/
```
aplanas, line 2194: what if there is no option, or is `none`? Uhm the original code take care of the first condition, but not of the second

sdbootutil
```diff
@@ -2158,33 +2176,31 @@ have_fido2()
...
-		[[ "$name" = \#* ]] && continue
-		if [[ "$opts" != *"$option"* ]]; then
-			[ -z "$opts" ] && opts="$option" || opts="$opts,$option"
-			# crypttab has changed so initrd needs to be updated
-			arg_no_reuse_initrd=1
-		fi
-		echo "$name $device ${key:-none} $opts" >> "$crypttab"
-	done < /etc/crypttab
+    # Use sed to add the option if it's not already present and line starts with cr_
+    sed -i -E "/^cr_/ {
```
aplanas, line 2191: editing in place can be dangerous, no?

sdbootutil
```diff
@@ -2158,33 +2176,31 @@ have_fido2()
...
+        s/(\S+\s+\S+\s+\S+\s*)(.*)/\1\2/
+        /$option/! {
+            s/$/,${option}/
+        }
+    }" "$crypttab_file"
 
-	mv -Z "$crypttab" /etc/crypttab
-	chmod 644 /etc/crypttab
+    # Check if the sed command made any changes
+    if ! cmp -s "${crypttab_file}" "${crypttab_file}.bak"; then
```
aplanas, line 2199: the bak file is kept?

aplanas:
I am experimenting with adding `x-sdbootutil.track` to mark the encrypted devices that should be tracked by sdbootutil (enroll, unenroll, etc), and seems to be a viable solution.

The problem that I see is the migration. Current systems should manually add this option in `/etc/crypttab` so the device is recognized by `sdbootutil`.  This is why I am thinking that maybe makes more sense to make is as an opt-out, so adding `x-sdbootutil.untrack` instead in the devices that `sdbootutil` should ignore.

What do you think?

aplanas:
https://github.com/openSUSE/sdbootutil/pull/153

TobiPeterG:
> I am experimenting with adding `x-sdbootutil.track` to mark the encrypted devices that should be tracked by sdbootutil (enroll, unenroll, etc), and seems to be a viable solution.
> 
> The problem that I see is the migration. Current systems should manually add this option in `/etc/crypttab` so the device is recognized by `sdbootutil`. This is why I am thinking that maybe makes more sense to make is as an opt-out, so adding `x-sdbootutil.untrack` instead in the devices that `sdbootutil` should ignore.
> 
> What do you think?

I'm slowly getting back on track and working through my different projects :)

Hmm as a user I wouldn't expect my drive to get enrolled, but only the root drive, but that might be me.
And now the other PR is already merged, so I guess we will go with this ;)
It's fine as well :)

aplanas:
> Hmm as a user I wouldn't expect my drive to get enrolled, but only the root drive, but that might be me.

My expectation is different, but I do not have a good argument to defend it.

But in any case cannot be the root drive only the one managed by sdbootutil. For example, cr_swap should be there too, and in MicroOS I guess etc and var (because of the overlays).  I am not sure about cr_home, but is hard for me to argue why sdbootutil should ignore it.

> And now the other PR is already merged, so I guess we will go with this ;)

We can change it

PR #161, opened 2024-11-25
sdbootutil-enroll: harden script against unexpected conditions

mgerstner:
- if CREDENTIALS_DIRECTORY is not set, then the script would use /$name, which could cause unexpected results. Therefore check for an existing variable first.
- in some situations a local user might be able to give arbitrary names to block devices. Thus only check the FSTYPE for crypto_LUKS, not the block device path.


aplanas (approved):

aplanas:
Just a nit: `&&` instead of `-a`. I will change it later.

Thanks, @mgerstner !!

PR #171, opened 2024-12-17
Use only the first output line of findmount in free space calc

Samonitari:


aplanas (approved):

aplanas:
Thanks @Samonitari! I will release a new version with your fix

Samonitari:
Very big thanks!

PR #175, opened 2024-12-20
Remove TUI

TobiPeterG:
Since @aplanas isn't a fan of the TUI (please correct me if I'm wrong) and the TUI doesn't have any functionality you can't do with the CLI, we can remove it. This will reduce the size of sdbootutil significantly, remove inconsistencies and make maintenance easier.

I tried to find every call to removed functions, please test thoroughly.

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -8,7 +8,6 @@ shopt -s nullglob
 # causes flicker
 dialog_altenate_screen=
 dialog_backtitle="sdbootutil"
-interactive=
 verbose=
 nl=$'\n'
 shimdir="/usr/share/efi/$(uname -m)"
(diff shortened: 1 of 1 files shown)


TobiPeterG:
Rebased

aplanas (approved):

aplanas:
There are some unused variables now: "dialog_backtitle", "dialog_altenate_screen" and the "--flicker" parameter,

TobiPeterG:
> There are some unused variables now: "dialog_backtitle", "dialog_altenate_screen" and the "--flicker" parameter,

Whoops, do you want to remove them or should I? :)

aplanas:
I will take care : ))

aplanas:
Should be in master now

PR #183, opened 2025-01-09
Mount /boot and subvolumes for chroot

aplanas:
Fix openSUSE/transactional-update#136


sdbootutil
```diff
@@ -613,26 +617,34 @@ mount_chroot()
 {
 	local snapshot_dir="$1"
 
-	# This is needed so that the rootfs appears in the mounts
-	# under the chroot, allowing dracut to properly detect the fs
-	# type and load the relevant module
+	# We include the rootfs (the first line usually), as is needed
+	# to appears in the mounts under the chroot, allowing
```
prawilny, line 621: nit: `s/appears/appear/`

sdbootutil
```diff
@@ -613,26 +617,34 @@ mount_chroot()
...
+		eval "$line"
+		[ "$FSTYPE" = "btrfs" ] || [ "$FSTYPE" = "vfat" ] || [ "$FSTYPE" = "xfs" ] || [[ "$FSTYPE" == ext* ]] || continue
+		[ "$TARGET" != "/" ] || continue
+		[[ "$TARGET" != /.snapshots* ]] || continue
+		mountpoint --quiet "$snapshot_dir$TARGET" || mount --bind "$TARGET" "$snapshot_dir$TARGET"
+	done < "$tmpdir/mounts"
+	rm "$tmpdir/mounts"
 
 	mount -t tmpfs -o size=10m tmpfs "$snapshot_dir/run"
-	for i in proc dev sys var tmp; do
```
prawilny, line 622: nit: is this removal needed?

sdbootutil
```diff
@@ -613,26 +617,34 @@ mount_chroot()
...
+	# to appears in the mounts under the chroot, allowing
+	# dracut to properly detect the fs type and load the relevant
+	# module.
+	findmnt -o TARGET,FSTYPE -Rv --pairs / > "$tmpdir/mounts"
 	mount --bind "$snapshot_dir" "$snapshot_dir"
+	while read -r line; do
+		eval "$line"
+		[ "$FSTYPE" = "btrfs" ] || [ "$FSTYPE" = "vfat" ] || [ "$FSTYPE" = "xfs" ] || [[ "$FSTYPE" == ext* ]] || continue
+		[ "$TARGET" != "/" ] || continue
+		[[ "$TARGET" != /.snapshots* ]] || continue
```
prawilny, line 630: Doesn't it mean that code below is always skipped?

prawilny (commented):
I can attest this code should fix the bug (I checked the diff against the commit I [reported the success with](https://github.com/openSUSE/transactional-update/issues/136#issuecomment-2594122631) and there were none around these changes).

sdbootutil
aplanas: done

sdbootutil
aplanas: yes, as var is now addressed by the iterator, if is a subvolume (that is one). `findmnt` will find it

sdbootutil
aplanas: no, `TARGET` is wrt the current system (/var, /root, /, /srv, etc). We do not want to mount `/.snapshots` or any other mount point inside it.  We want to replicate the current system in the chroot

aplanas:
@prawilny thanks for working with me in this issue. I will try to release a new version soon : )

PR #187, opened 2025-01-10
add info about how to ignore devices to --help

528-491:
Hello,

this is my first ever PR, please be kind :)

It simply adds information to the --help invocation about how to ignore LUKS2 devices you don't want to manage with sdbootutil.

It helped me when dealing with multiple encrypted OS installations on my machine and is a simple copy/paste from [a comment](https://github.com/openSUSE/sdbootutil/blob/f762ec762f805a93484211f228837d5c51f9747d/sdbootutil#L2107) inside the `detect_tracked_devices()` function.

Thanks!


aplanas (approved):
Thanks a lot for the commit. Makes a lot of sense.

PR #188, opened 2025-01-10
Accept spaces or tabs when parting crypttab

aplanas:


TobiPeterG:
Just tested, this works as well :)

PR #191, opened 2025-01-15
Regenerate initrd if missing

### LINKED ISSUE #136: Regenerate missing initrd
When adding a kernel a missing initrd should be regenerated. Similar to the option `--no-reuse-initrd`.

In my case I had space problems on my efi partition so I removed old kernels with zypper and also removed all folders from `/boot/efi/long-random-string`. I had assumed that everything needed would be regenerated with `sdbootutil add-all-kernels` but this time it only added the kernels without the initrds.

aplanas:
Fix #136

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -599,6 +599,7 @@ reuse_initrd()
 		local k v
 		while read -r k v; do
 			[ "$k" = 'initrd' ] || continue
+			[ -f "${boot_root}$v" ] || continue
 			log_info "found existing initrd $v"
 			dstinitrd+=("$v")
 		done < "$conf"
@@ -1652,7 +1653,7 @@ pcrlock_manual_raw()
 {
 	local pcr="$1"
 	local pcrlock="$2"
-	local initrd="$3"
+	local element="$3"
 
 	echo -n '{"records":[{"pcr":'"$pcr"',"digests":[' > "$pcrlock"
 	local separator=0
@@ -1661,7 +1662,7 @@ pcrlock_manual_raw()
 		[ "$separator" = "0" ] || echo -n "," >> "$pcrlock"
 		separator=1
 		hash "${a}sum" || continue
-		read -r digest _ < <("${a}sum" "$initrd")
+		read -r digest _ < <("${a}sum" "$element")
 		echo -n '{"hashAlg":"'"$a"'","digest":"'"$digest"'"}' >> "$pcrlock"
 	done
 	echo ']}]}' >> "$pcrlock"
@@ -1797,12 +1798,16 @@ pcrlock_sdboot()
 	# pcrlock standards
 	# TODO: move to kernel-TYPE-pcrlock.rpm
 	local n=0
-	while read -r i; do
+	while read -r linux; do
+		[ -f "${boot_root}$linux" ] || {
+			log_info "Missing ${boot_root}$linux, ignoring entry for prediction"
+			continue
+		}
 		n=$((n+1))
 		pcrlock \
 			lock-pe \
 			--pcrlock=/var/lib/pcrlock.d/650-kernel-efi-application.pcrlock.d/linux-"$n".pcrlock \
-			"${boot_root}/$i"
+			"${boot_root}/$linux"
 	done < <(jq --raw-output 'map(.linux) | unique | .[]' "$entryfile")
 
 	# Join the cmdline and the initrd in a single component
@@ -1810,6 +1815,10 @@ pcrlock_sdboot()
 	n=0
 	while read -r cmdline; do
 		read -r initrd
+		[ -f "${boot_root}$initrd" ] || {
+			log_info "Missing ${boot_root}$initrd, ignoring entry for prediction"
+			continue
+		}
 		n=$((n+1))
 		pcrlock_sdboot_cmdline_initrd "initrd=$cmdline" "${boot_root}$initrd" "$n"
 	done < <(jq --raw-output '.[] | ([(.initrd[0] | sub("/"; "\\"; "g")), .options] | join(" ")), .initrd[0]' "$entryfile"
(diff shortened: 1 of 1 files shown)


TobiPeterG:
How can I test this? Does this only occur on systems using GRUB?

aplanas:
it is fairly easy to reproduce, and happens in systemd-boot and grub. Remove some initrd and call add-kernel (or add-all-kernels) for the snapshot that this initrd belong to. With the bug the initrd will remain removed and some error messages (from pcrlock) will appear. Now the initrd should be created for the selected snapshot.

I tested this, and so far looks good.  Maybe you want to evaluate the code.

sdbootutil
```diff
@@ -1797,19 +1798,21 @@ pcrlock_sdboot()
 	# pcrlock standards
 	# TODO: move to kernel-TYPE-pcrlock.rpm
 	local n=0
-	while read -r i; do
+	while read -r linux; do
+		[ -f "${boot_root}$linux" ] || continue
```
TobiPeterG, line 1802: Wouldn't it be useful to regenerate the entry when we detect that the initrd is missing?

sdbootutil
```diff
@@ -599,6 +599,7 @@ reuse_initrd()
 		local k v
 		while read -r k v; do
 			[ "$k" = 'initrd' ] || continue
+			[ -f "${boot_root}$v" ] || continue
```
TobiPeterG, line 602: Same here, this meams that the parent's initrd is missing, right?
Shouldn't we regenerate the parent's entry?

TobiPeterG: (same for every other place where we detect that the initrd is missng)

TobiPeterG (commented):
Looks good, nice that you rename generic variables to something more useful :)

Just thinking about just ignoring that the intitrd is missing while measuring.
If we already detect this case, we can also fix it IMO.

sdbootutil
aplanas: I do not think so. `sdbootutil` removes entries when it want to free space (so calling the garbage collector via `bootctl cleanup` will remove the kernel and / or the initrd if they are not used anymore).  If the initrd or the kernel is missing then is because the user removed it, for some reason, breaking the boot entry. 

Once it is broken, there is no good decision, except when the user re-add the kernel (add-kernel, so recreate the initrd, that is the goal of this PR) or do a new prediction (skip predictions of broken entries, that it was is done here too)

Recreating the entry, or remove the entry will be a surprise for the user.

sdbootutil
aplanas: maybe, but I decided that we should not affect the content of other snapshots. So if the current initrd is missing and the parent too, then it will create the one for the current snapshot. The user can create the one for the parent passing the SNAPSHOT parameter

sdbootutil
TobiPeterG: What would be a reason to have a broken entry? In which case would that be useful and expected by the user?
I guess if we don't want to regenerate it, we should at least remove the entry, as it is in fact useless :D
We can print a warning that we removed a broken entry.

As user, I would wish to not have broken entries in my boot menu and for sdbootutil to detect these situations or at least react accordingly.

aplanas: > What would be a reason to have a broken entry? In which case would that be useful and expected by the user?

The user can do a `rm initrd-a4f*` in the ESP, for example. I do that when developing and testing. It is not usual, but sometimes removing an initrd can help to debug issues with included modules, free some space for a moment, etc.

> I guess if we don't want to regenerate it, we should at least remove the entry

This code is in the prediction. It is called from different places, for example `sdbootutil update-predictions`.  If I removed one initrd I broke one entry, I do not want to regenerate the initrd during the update of predictions, nor removing the entry.  The sensible thing to do is ignore them when updating the predictions.  Note that the broken entry could be anyone, not the default one.  Can be from a very old snapshot.

> as it is in fact useless :D

It is. But you can generate the initrd back with `sdbootutil mkinitrd` and now (with this patch) with `sdbootutil add-all-kernel`.  With the current patch you will regenerate all the missing initrds for the current snapshot.

> We can print a warning that we removed a broken entry.

During the "update-predictions" the system should not be altered.

> or at least react accordingly

But this is what this patch is doing? Re-add the missing initrds when calling add-all-kernels, and ignore the broken entries when update-predictions. I can add a warning during the update-prediction to point the broken entries, would that help?

TobiPeterG: > > What would be a reason to have a broken entry? In which case would that be useful and expected by the user?
> 
> The user can do a `rm initrd-a4f*` in the ESP, for example. I do that when developing and testing. It is not usual, but sometimes removing an initrd can help to debug issues with included modules, free some space for a moment, etc.
> 

Oh alright, I didn't have a use case yet where I wanted to only remove the inird. So you also need the entry config file after deleting the initrd? If yes, just out of curiosity, what for? If not, why would it be bad to remove it?

> > I guess if we don't want to regenerate it, we should at least remove the entry
> 
> This code is in the prediction. It is called from different places, for example `sdbootutil update-predictions`. If I removed one initrd I broke one entry, I do not want to regenerate the initrd during the update of predictions, nor removing the entry. The sensible thing to do is ignore them when updating the predictions. Note that the broken entry could be anyone, not the default one. Can be from a very old snapshot.
> 

Yes, it can be from an old snapshot, but what does it change? The entry is not bootable. A user might be wondering why it doesn't boot and doesn't know what's wrong. What value does the broken entry provide that we want to keep?

> 
> During the "update-predictions" the system should not be altered.
> 

Fair, but removing something that's not working anyway doesn't really feel like "altering the system", though I see what you mean.

> > or at least react accordingly
> 
> But this is what this patch is doing? Re-add the missing initrds when calling add-all-kernels, and ignore the broken entries when update-predictions. I can add a warning during the update-prediction to point the broken entries, would that help?

A warning would definitely be better than nothing ;)

I'm curious: When you remove an initrd or have a broken initrd on your system and you call sdbootutil ..., be it add-all-kernels or update-predictions, do you expect that sdbootutil detects broken entries and removes them? I mean, isn't the purpose of sdbootutil to manage the boot entries and initrds of a system? Personally, I would expect sdbootutil to "correct" my system, with every entry shown on next boot working.
When a user messes with what sdbootutil is there for, I think it's his responsibility to not call sdbootutil to keep this "broken" setup.
But, as I said, just my opinion.
I'm just afraid of users having unbootable snapshot entries, not knowing about sdbootutil and not knowing why or what to do.

aplanas: > So you also need the entry config file after deleting the initrd? If yes, just out of curiosity, what for? If not, why would it be bad to remove it?

Yes, because is a transient situation. The entry is the important stuff, I remove the initrd to make some test (recreate it, make some free space). I want to go back to the original state (recreate the initrd), and with this patch we can do that with mkinitrd and add-kernel.

> A user might be wondering why it doesn't boot and doesn't know what's wrong. What value does the broken entry provide that we want to keep?

True. This patch is not about answering this question, but about not erroring when the user wants to fix the state (re-create the initrd). In any case sdbootutil making the decision of removing the entry is never the good approach.

> A warning would definitely be better than nothing ;)

I agree. I added a message now.

> do you expect that sdbootutil detects broken entries and removes them?

No. If I removed the initrd I should be able to do "mkinitrd" for this entry later. If there is a process that removes this entry without my intervention (for example, after calling update-predictions), I cannot later do the call to mkinitrd, as the entry is not there anymore.

> I would expect sdbootutil to "correct" my system, with every entry shown on next boot working.

Sure, maybe what is missing is a new command that remove broken entries? Like `sdbootutil cleanup`, that calls `bootctl cleanup` to remove unused kernel and initrd, and also remove later the broken entries.

PR #196, opened 2025-01-27
Switch to nested subvolume based /etc

laenion:
transactional-update 5.0.0 is moving away from the complex overlayfs based system for /etc to a nested subvolume based system. This also makes a lot of the existing code unnecessary.

This pull request adapts the code accordingly - as mentioned large portions of the code don't seem to be necessary any more. In my personal testing it seemed to work correctly (though I couldn't get the initrd to be regenerated), but I certainly don't know all the subtle details of the script I may have missed.

Test packages for _transactional-update_ and _read-only-root-fs_ which contain the relevant changes for the new /etc handling can be found in https://build.opensuse.org/project/show/home:fos:branches:Base:System.

(An alternative to check for a transactional system would be to use `snapper list --columns default,number,read-only` instead of the rpm command (see internal https://suse.slack.com/archives/C02BPCB2U1M/p1693400438485489 discussion for more ideas), but I think this is short and concise.)

### DIFF
--- a/10-sdbootutil.snapper
+++ b/10-sdbootutil.snapper
@@ -6,7 +6,7 @@ set -e
 # check whether it's a transactional system
 is_transactional()
 {
-	[ "$(stat -f -c %T /etc)" = "overlayfs" ]
+	rpm --query --quiet read-only-root-fs
 }
 
 # when creating a snapshot we fetch all bls configs from previous snapshot dir,
--- a/sdbootutil
+++ b/sdbootutil
@@ -278,9 +278,7 @@ reset_rollback()
 # check whether it's a transactional system
 is_transactional()
 {
-	# For running systems we can have this tests instead:
-	#   [ "$(stat -f -c %T /etc)" = "overlayfs" ]
-	grep -q "^overlay /etc" /etc/fstab
+	rpm --query --quiet read-only-root-fs
 }
 
 keyctl_add_with_timeout()
@@ -570,20 +568,10 @@ set_machine_id()
 	local snapshot="$1"
 	local subvol=""
 	[ -z "$snapshot" ] || subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
-	machine_id_files=()
-	if is_transactional && [ -z "$TRANSACTIONAL_UPDATE" ]; then
-		[ -n "$snapshot" ] && machine_id_files+=("/var/lib/overlay/$snapshot/etc/machine-id")
+	machine_id_file="${subvol#"${subvol_prefix}"}/etc/machine-id"
+	if [ -s "$machine_id_file" ]; then
+		read -r machine_id < "$machine_id_file"
 	fi
-	machine_id_files+=(
-		"${subvol#"${subvol_prefix}"}/etc/machine-id"
-	)
-
-	for file in "${machine_id_files[@]}"; do
-		if [ -s "$file" ]; then
-			read -r machine_id < "$file"
-			break
-		fi
-	done
 }
 
 reuse_initrd()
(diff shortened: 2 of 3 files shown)


aplanas:
@laenion thanks a lot!

I am wondering about the transition from one model to another.  Would make sense to have both methods in sdbootutil for a while?

TobiPeterG:
> @laenion thanks a lot!
> 
> I am wondering about the transition from one model to another.  Would make sense to have both methods in sdbootutil for a while?

Hmm maybe rather define in the spec that sdbootutil conflicts with transactional-update< 5.0.0
A system can only have one tu version installed, either the old one or the new one, we just have to prevent that on a system with the old tu, we install sdbootutil in a new version.
Additionally, tu 5.0 should conflict with sdbootutil< the first version that supports tu 5.0.0 to prevent the opposite issue.
That would be my solution :)

aplanas:
> sdbootutil conflicts with transactional-update< 5.0.0

Ok for me.

> A system can only have one tu version installed

I am not sure. tu will download the rpm of the new version and use it without installing it. So you will use 5.0 before it is installed.

@laenion does tu 5.0 takes care of migrating the `/etc` model during the update?

> (though I couldn't get the initrd to be regenerated)

`sdbootutil mkinitrd` should do the trick, but I guess this needs to be tested more carefully.

Do we need to sync the release?

laenion:
> > sdbootutil conflicts with transactional-update< 5.0.0
> 
> Ok for me.

And transactional-update will conflict with sdbootutil in whatever version this change will be contained in :)

> > A system can only have one tu version installed
> 
> I am not sure. tu will download the rpm of the new version and use it without installing it. So you will use 5.0 before it is installed.

There's a new check that will prevent using the new version if there's a major release in between. The reason is that the new version requires a lot more files, e.g. the snapper plugin, which can't just be downloaded and used immediately.

> @laenion does tu 5.0 takes care of migrating the `/etc` model during the update?

Yes, migration to the new model is automatic.

> > (though I couldn't get the initrd to be regenerated)
> 
> `sdbootutil mkinitrd` should do the trick, but I guess this needs to be tested more carefully.
> 
> Do we need to sync the release?

Yes, we will have to - the new sdbootutil will require the new transactional-update and vice versa.

aplanas:
> And transactional-update will conflict with sdbootutil with versions lower than the one this change will be contained in :)

Do you mind to add the Conflict: transactional-update < 5.0 [here](https://github.com/openSUSE/sdbootutil/blob/main/sdbootutil.spec) as part of the same PR?

laenion:
Done!

aplanas (approved):
LGTM. Before merge I should test it a bit better.

10-sdbootutil.snapper
```diff
@@ -6,7 +6,7 @@ set -e
 # check whether it's a transactional system
 is_transactional()
 {
-	[ "$(stat -f -c %T /etc)" = "overlayfs" ]
+	rpm --query --quiet read-only-root-fs
```
Vogtinator, line 9: I think we can check for `ro` in `/etc/fstab` instead: `findmnt --fstab / -O ro >/dev/null`

sdbootutil
```diff
@@ -570,20 +568,10 @@ set_machine_id()
 	local snapshot="$1"
 	local subvol=""
 	[ -z "$snapshot" ] || subvol="${subvol_prefix}/.snapshots/${snapshot}/snapshot"
-	machine_id_files=()
-	if is_transactional && [ -z "$TRANSACTIONAL_UPDATE" ]; then
-		[ -n "$snapshot" ] && machine_id_files+=("/var/lib/overlay/$snapshot/etc/machine-id")
+	machine_id_file="${subvol#"${subvol_prefix}"}/etc/machine-id"
```
Vogtinator, line 571: We still need to support snapshots with overlays I think, so let's not touch this

Vogtinator:
Superseded by #242

PR #198, opened 2025-02-04
Measure the volume key of tracked devices in PCR 15

aplanas:


aplanas:
> I had a brief look, what happens when the PCR value doesn't fit the measured one?

https://github.com/openSUSE/sdbootutil/pull/198/files#diff-d8e626c7bfe88c983f67b7a9b9ed2d9ba89788f9e4eb041b46cd2a2b6323a754R5

> Also should the dracut module really live here?

Can be a different project I guess. In dracut-pcr-signature, a new one?

TobiPeterG:
> > I had a brief look, what happens when the PCR value doesn't fit the measured one?
> 
> https://github.com/openSUSE/sdbootutil/pull/198/files#diff-d8e626c7bfe88c983f67b7a9b9ed2d9ba89788f9e4eb041b46cd2a2b6323a754R5
> 
> > Also should the dracut module really live here?
> 
> Can be a different project I guess. In dracut-pcr-signature, a new one?

Ah exit 1, missed that.
This will mean that the user ends up in the emergency shell, right? Is this really the sane way? Normally, when a PCR doesn't match, the TPM just won't provide the passphrase. Would it be possible to instead just "lock" the TPM and continue with the boot? Afterall, nothing has been executed from the compomised partition yet.

TobiPeterG:
> > Ah exit 1, missed that.
> 
> > emergency shell, right?
> 
> The validator script does that, but the service will call systemd-halt.service, so no emergency shell.
> 
> > Would it be possible to instead just "lock" the TPM and continue with the boot? Afterall, nothing has been executed from the compomised partition yet.
> 
> This option is contemplated, yes. Currently the system will halt, stopping the initrd to do the switch root into the compromised device (you will never know what kind of exploit is waiting). But the user can add in the cmdline "measure-pcr-validator.ignore=yes", invalidating the TPM2 prediction and the validator will log a warning and continue the boot.

Hmm I guess as a user I would be very confused that my PC that I just turned on turned itself back off :D 
What would be against just locking the TPM? Every other component could be compromised and the boot would continue, so why should it be special for this one?

aplanas:
> Hmm I guess as a user I would be very confused that my PC that I just turned on turned itself back off :D

Uhmm yes, I agree. I hope this can be documented and explained. For sure we need to iterate more in UX.

But for now, IMHO, it is doing the right thing: keep the user safe.

One improvement that I can think of is to present a visible message of the situation and the cmdline option that bypass it.

> What would be against just locking the TPM?

You mean invalidating the TPM2 PCRs, extending all of them? It is already too late!! The device was unlocked!!

> Every other component could be compromised and the boot would continue, so why should it be special for this one?

"systemd-cryptsetup" check the "tpm2-measure-pcr=yes" to extend PCR 15 with the volume key. To get the volume key it needs a password, so is using the one from the TPM2, so the unlock is happening before the PCR 15 is extended. In other words, the PCR 15 cannot participate in the policy.

So the PCR 15 validation is happening too late, when all the devices are already open, and the next step is the switch root.  The invalidation of the TPM2 policy needs to happen before, so this is why the "measure-pcr-validator.ignore=yes" is there, to invalidate the policy and force to ask the password (and bypass the halt)

TobiPeterG:
> > Hmm I guess as a user I would be very confused that my PC that I just turned on turned itself back off :D
> 
> Uhmm yes, I agree. I hope this can be documented and explained. For sure we need to iterate more in UX.
> 
> But for now, IMHO, it is doing the right thing: keep the user safe.
> 
> One improvement that I can think of is to present a visible message of the situation and the cmdline option that bypass it.
> 
> > What would be against just locking the TPM?
> 
> You mean invalidating the TPM2 PCRs, extending all of them? It is already too late!! The device was unlocked!!
> 
> > Every other component could be compromised and the boot would continue, so why should it be special for this one?
> 
> "systemd-cryptsetup" check the "tpm2-measure-pcr=yes" to extend PCR 15 with the volume key. To get the volume key it needs a password, so is using the one from the TPM2, so the unlock is happening before the PCR 15 is extended. In other words, the PCR 15 cannot participate in the policy.
> 
> So the PCR 15 validation is happening too late, when all the devices are already open, and the next step is the switch root. The invalidation of the TPM2 policy needs to happen before, so this is why the "measure-pcr-validator.ignore=yes" is there, to invalidate the policy and force to ask the password (and bypass the halt)

Hmm I'm not quite sure if I understand the concerns, sorry if something I say now is wrong/invalid:

Why is it too late to lock the TPM when the device has been unlocked? In the attack scenario I know, the TPM isn't enrolled to the drive anyway, so it can't provide the passphrase and the attacker has to enter his passphrase manually. Afterwards, when the compromized system is booted, he can ask the TPM "Hey, give me your password" and the TPM provides it, as every PCR measurement was successful.
What my idea would be:
We detect that PCR 15 doesn't match, so we lock the TPM. Yes, the compromized drive has already been unlocked, but the attacker doesn't need the TPM for this anyway.
Now, we switch root into the compromized system. The attacker asks the TPM "Hey, give me your password", but now the TPM says "no", as we locked it. The attacker can't get the passphrase, the attack failed.
Where is my thought process wrong?

aplanas:
> Why is it too late to lock the TPM when the device has been unlocked?

Please, clarify what do you mean for "lock the TPM2".  I think we can be talking about different stuff.

TobiPeterG:
> > Why is it too late to lock the TPM when the device has been unlocked?
> 
> Please, clarify what do you mean for "lock the TPM2". I think we can be talking about different stuff.

Ah.
With "locking TPM" I mean either: The TPM won't provide the passphrase for this boot cycle (as if a measurement was wrong (which it is))
or
Lock it forever, so it won't provide the passphrase even if the correct partition was reintroduced, so a new enrollment is neccessary

I thought at least one of these two should be possible :)

aplanas:
> The TPM won't provide the passphrase for this boot cycle (as if a measurement was wrong (which it is))

It is not. The policy is build with other PCRs except 15. As commented, the PCR 15 extension done by systemd is happening after the policy is validated. The policy does not include PCR 15. So in this context, blocking the TPM2 means extend all the PCRs from 0..=9 (maybe even to 12). This can be done, yes.

> Lock it forever, so it won't provide the passphrase even if the correct partition was reintroduced, so a new enrollment is neccessary

This is done by a configuration option in the TPM2. Is a dictionary attack lock-down (DA protection), and is configured per TPM2. IFAIK I cannot trigger it unless it is configured and I retry to open an invalid policy multiple times (https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_dictionarylockout.1.md)

For now, I will understand "locking TPM2" for the first assertion: moving the PCRs in a situation that the current policy is invalid.

> In the attack scenario I know [...]

Agreed.

> What my idea would be:
> We detect that PCR 15 doesn't match, so we lock the TPM [...]
> Now, we switch root into the compromized system. The attacker asks the TPM "Hey, give me your password", but now the TPM says "no", as we locked it. The attacker can't get the passphrase, the attack failed.
Where is my thought process wrong?

Understood. Instead of doing the halt we extend all the PCRs from 0 to 9 or 12, and now is not possible to use the policy to extract the password associated for the TPM2 key slot. That works, but are still giving access to a rogue device that can do anything.

In the more general case we are not sure of the kind of exploit that can be used. You could exploit the kernel using specifically designed block devices, that can be used to dump the memory, from a system that already recovered the TPM2 password.

But what is worse, if you have multiple devices, the initrd will open those other devices. Even invalidating the TPM2 policy, the rogue device will have access to the data of those other devices!! As commented, it is too late.

TobiPeterG:
> > The TPM won't provide the passphrase for this boot cycle (as if a measurement was wrong (which it is))
> 
> It is not. The policy is build with other PCRs except 15. As commented, the PCR 15 extension done by systemd is happening after the policy is validated. The policy does not include PCR 15. So in this context, blocking the TPM2 means extend all the PCRs from 0..=9 (maybe even to 12). This can be done, yes.
> 
> > Lock it forever, so it won't provide the passphrase even if the correct partition was reintroduced, so a new enrollment is neccessary
> 
> This is done by a configuration option in the TPM2. Is a dictionary attack lock-down (DA protection), and is configured per TPM2. IFAIK I cannot trigger it unless it is configured and I retry to open an invalid policy multiple times (https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_dictionarylockout.1.md)
> 
> For now, I will understand "locking TPM2" for the first assertion: moving the PCRs in a situation that the current policy is invalid.
> 
> > In the attack scenario I know [...]
> 
> Agreed.
> 
> > What my idea would be:
> > We detect that PCR 15 doesn't match, so we lock the TPM [...]
> > Now, we switch root into the compromized system. The attacker asks the TPM "Hey, give me your password", but now the TPM says "no", as we locked it. The attacker can't get the passphrase, the attack failed.
> Where is my thought process wrong?
> 
> Understood. Instead of doing the halt we extend all the PCRs from 0 to 9 or 12, and now is not possible to use the policy to extract the password associated for the TPM2 key slot. That works, but are still giving access to a rogue device that can do anything.
> 
> In the more general case we are not sure of the kind of exploit that can be used. You could exploit the kernel using specifically designed block devices, that can be used to dump the memory, from a system that already recovered the TPM2 password.
> 
> But what is worse, if you have multiple devices, the initrd will open those other devices. Even invalidating the TPM2 policy, the rogue device will have access to the data of those other devices!! As commented, it is too late.


Ah aha, the multiple disks issue is a concern I didn't think about.
I guess locking the non-compromized partitions is not feasible, so yes, we should prevent booting into the compromised system.

Would it be possible to, instead of halting the system, just print a big fat red warning like: "The system drive is not in the expected state. (PCR15)"
And maybe a message on how to prevent this message with the cmdline option, but not turn off the system, but stay in this state?
Is it possible to set the cmdline of the next boot to automatically set the option next time the system is booted 
We could delete the pcrlock file on the efi partition so the TPM won't unlock - anything on the next boot.
Maybe even do this by default, just delete the pcrlock file and reboot? Would this already prevent the attack, but still allow the user to boot when the PCR 15 doesn't match for a non-attack reason? He would then have to enter the passphrase for his system and drives (that the user knows) and an attacker can't access any drive that was unlocked by the TPM as the TPM won't unlock any drive without a pcrlock file.

aplanas:
> Would it be possible to, instead of halting the system, just print a big fat red warning like: "The system drive is not in the expected state. (PCR15)"

"Instead" no. You need to halt the system, I do not think that it is optional? But yes, a message is useful here. I agree. I will research how to write this kind of messages, as systemd is collecting the output for the journal, and I will update the PR.

> And maybe a message on how to prevent this message with the cmdline option, but not turn off the system, but stay in this state?

Yes, I will add info about the cmdline option in the message. Note that the halt does not shutdown the system. It is freezing it.

> We could delete the pcrlock file on the efi partition so the TPM won't unlock - anything on the next boot.
> Maybe even do this by default, just delete the pcrlock file and reboot? Would this already prevent the attack, but still allow the user to boot when the PCR 15 doesn't match for a non-attack reason? He would then have to enter the passphrase for his system and drives (that the user knows) and an attacker can't access any drive that was unlocked by the TPM as the TPM won't unlock any drive without a pcrlock file.

That does not fix anything ... and is not adding anything on top of the current approach. Or what am I missing? There is nothing to fix for the next boot ... the issue is the current one.

aplanas:
> print a big fat red warning like: "The system drive is not in the expected state. (PCR15)"

Done. In white.

TobiPeterG:
> > print a big fat red warning like: "The system drive is not in the expected state. (PCR15)"
> 
> Done. In white.

Perfect, thanks :)

Is there an easy way to test this? :D

aplanas:
> Is there an easy way to test this? :D

* Download `pcr-signature.sh` from https://github.com/aplanas/dracut-pcr-signature
* `cp pcr-signature.sh /usr/lib/dracut/modules.d/50pcr-signature/.`
* Download the new code from this PR
* `chmod a+x *.sh; cp -a measure* module* /usr/lib/dracut/modules.d/50measure-pcr/.`
* Add `x-initrd.attach,tpm2-device=auto,tpm2-measure-pcr=yes` as options in `/etc/crypttab` for the devices that should be in `initrd`

When enrolling a public and private keys are generated, so the prediction can be measured, so:
* `mkdir -p /var/lib/sdbootutil`
* `chmod 700 /var/lib/sdbootutil`
* `openssl genrsa -out /var/lib/sdbootutil/measure-pcr-private.pem 4096`
* `openssl rsa -in /var/lib/sdbootutil/measure-pcr-private.pem -pubout -out /var/lib/sdbootutil/measure-pcr-public.pem`

Now `sdbootutil update-predictions` should create the file `measure-pcr-prediction` and the signature `measure-pcr-prediction.sha256` in `/var/lib/sdbootutil` and copy then in `/boot/efi/EFI/systemd`

You can test that if the file is there but wrong, `update-predictions --mesure-pcr` will recreate it with the correct values. If there are multiple disk, the order on how they are measure is always the same (no race condition), if the signature is not valid, the `initrd` halt, or than `measure-pcr-validator.ignore=1` will bypass the check, but will ask for the password to unlock the device. And many other things

aplanas:
Documented here: https://en.opensuse.org/Portal:MicroOS/FDE#PCR_#15

aplanas:
@TobiPeterG Did you have a chance to test it? I fixed some bugs but so far the code seems stable.

TobiPeterG:
> @TobiPeterG Did you have a chance to test it? I fixed some bugs but so far the code seems stable.

Not yet, will do so later today :)

aplanas:
@TobiPeterG Would be OK to merge it today, and fix issues later?

TobiPeterG:
> @TobiPeterG Would be OK to merge it today, and fix issues later?

Hey, yes, sorry
I was busy this week with exams and couldn't do everything I planned. Sorry
I wanted to test this today, would this be ok: If I'm not responding until today evening, you just merge it? :)

aplanas:
> sorry

No need to! I still remember how hard are end of January and February.

> I wanted to test this today

Sure. In any case I expect issues. I tested it for several days but I am sure there are evident ways to improve it. If I merge it we can fix in subsequent requests.

PR #212, opened 2025-02-27
Fix version compare and typo

radolin:
This PR includes a fix from Michael to the version compare function between the bootloader installed in the ESP and the system one, which seems to be backwards. I also noticed a typo in a variable name and updated the help text.

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -164,14 +164,16 @@ helpandquit()
 		           is potentially bootable
 
 		install
-		           Install systemd-boot and shim into ESP
+		           Install the bootloader and shim into ESP
 
 		needs-update
 		           Check whether the bootloader in ESP needs updating
 
 		update
-		           Update the bootloader if it's older than the deployed
-		           version.  Use the --sync to allow downgrades
+		           Update the bootloader in the ESP if a newer version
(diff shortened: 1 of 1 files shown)


sdbootutil
```diff
@@ -1272,17 +1274,17 @@ bootloader_needs_update()
...
 	bldr_name=$(bootloader_name "$snapshot")
 	if [ "$status" = "11" ]; then
-		log_info "$bldr_name needs to be updated"
-		return 0
-	elif [ "$status" = "12" ]; then
 		log_info "$bldr_name is newer than system bootloader"
 		return 2
+	elif [ "$status" = "12" ]; then
+		log_info "$bldr_name needs to be updated"
+		return 0
```
aplanas, line 1287: I think that this is wrong.

  n = version of the current snapshot (that should be the version in the ESP)
  nv = version of the older snapshot

With `--sync` we support downgrades, then `v > nv` and so `systemd-analyze compare-versions $v $nv` will retcode `11`

aplanas: Ahhh ... I see now. Later is compared with `[ $status -eq 2 ]` if `--sync`, so `11` in the new code.

aplanas (approved):

PR #242, opened 2025-04-09
Support transactional systems without overlayfs on /etc

Vogtinator:
Alternative to #196 with backwards-compat


sdbootutil
```diff
@@ -326,9 +326,7 @@ reset_rollback()
 
 is_transactional()
 {
-	# For running systems we can have this tests instead:
-	#   [ "$(stat -f -c %T /etc)" = "overlayfs" ]
-	grep -q "^overlay /etc" /etc/fstab
+	findmnt --fstab / -O ro >/dev/null
```
aplanas, line 329: Is this always true? I though that there were a version of TW (transactional server or something) that has the traditional layout but in a RO mode. If this is true, this will confuse the code making it believe that is a MicroOS.

Also I think that if TW with a btrfs in save mode is booted, the / is also RO?

Vogtinator:
Tested successfully in all t-u combinations: old t-u, new t-u with old layout and new t-u with new layout.

Issue #168, opened 2024-12-09
[feature] Support encryped LVM partitions


albundy83:
Hello,

I have followed the guide from [here](https://news.opensuse.org/2024/09/20/quickstart-fde-yast2/)
But when the enroll step comes with:
```bash
gbellongervais@localhost:~> sudo -i
gbellongervais@localhost:~> sudo sdbootutil enroll --method tpm2
[sudo] password for root: 
Garbage after device path end, ignoring.
Garbage after device path end, ignoring.
Disk does not have GPT partition table, refusing.
```

But as you can see, I have a GPT partition table:
```bash
gbellongervais@localhost:~> sudo fdisk -l
Disk /dev/nvme0n1: 476.94 GiB, 512110190592 bytes, 1000215216 sectors
Disk model: PM9A1 NVMe Samsung 512GB                
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt       # <--- my GPT partition table
Disk identifier: 7083F1BE-877F-4A53-AAF7-7B271323C0F9

Device           Start        End   Sectors   Size Type
/dev/nvme0n1p1    2048    1050623   1048576   512M EFI System
/dev/nvme0n1p2 1050624 1000215182 999164559 476.4G Linux LVM


Disk /dev/mapper/cr_nvme-eui.36483330541661670025385800000001-part2: 476.42 GiB, 511555476992 bytes, 999131791 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes


Disk /dev/mapper/system-swap: 2 GiB, 2147483648 bytes, 4194304 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes


Disk /dev/mapper/system-root: 474.42 GiB, 509406609408 bytes, 994934784 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
```

Here another output for disk structure:
```bash
gbellongervais@localhost:~> sudo lsblk 
[...]

PR #255, opened 2025-05-09
Measure GPT of the disk with ESP, not the disk with root

arvidjaar:
Firmware measures GPT of the disk where boot application was loaded from.
In practice it means ESP where we install our bootloader.

systemd-pcrlock defaults to the disk where root ("/") filesystem is
located. It can be anywhere and it can be on a virtual device, like
LUKS container or LVM, but it is not what firmware knows or measures.

Explicitly measure GPT with ESP to match the firmware behavior.

Fixes: #168


aplanas (approved):

aplanas:
Thanks @arvidjaar

Issue #263, opened 2025-07-01
sdbootutil doesn't seem to work if I run "sudo transactional-update dup"


LarryIsBetter:
I'm sorry if this isn't the correct place to report this please tell me if it isn't.

[systemdupdate.txt](https://github.com/user-attachments/files/21005982/systemdupdate.txt)
[transactionaldup.txt](https://github.com/user-attachments/files/21005981/transactionaldup.txt)

Running "sudo journalctl | grep sdbootutil" after running  "sudo transactional-update dup" produces no new logs so I assume it isn't working

But if I check the journalctl logs after running "sudo systemctl start transactional-update.service" the logs start getting filled up.

I believe this issues causes [this](https://bugzilla.opensuse.org/show_bug.cgi?id=1244451).

People have been working around this issue with [this](https://www.reddit.com/r/AeonDesktop/comments/1koa394/comment/msorfdy/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button).

With them running "sdbootutil -v add-all-kernels" it further leads me to believe this is a sdbootutil issue?

LarryIsBetter:
I think I almost broke my system trying to do that. (not your fault I'm just an idiot)

The log file was empty for both the update methods, I think I did something wrong and I don't want to try it again.

If the sdbootutil plugin is called and sdbootutil isn't being triggered then I assume it's not an issue that should belong here, but be reported to the person who manages the plugin correct?

If you can, can you link me to where who I should report the issue?

aplanas:
> If the sdbootutil plugin is called and sdbootutil isn't being triggered then I assume it's not an issue that should belong here, but be reported to the person who manages the plugin correct?

The plugin is in this project, so I should be able to help.  I do not use Aeon, but from your messages I understand that this error only triggers when kernel is updated? Is there for me a reliable way to reproduce the error?

LarryIsBetter:
Ok correct me if I'm wrong.

From what I understand the sdbootutil plugin should trigger every time an update has been triggered, doesn't matter if the user manually triggered it or the auto update service did right?

If that's correct it seems that it doesn't just happen when the kernel is updated but every time a user runs "sudo transactional-update dup".

It's most noticeable after a kernel update though because people see that their networking, audio, USB ect ect don't work. So when they start looking around for the issue and they see kernel version isn't the one it's meant to be.

[...]

PR #264, opened 2025-07-07
Check for transactional systems by actually checking ro status

sysrich:
Checking the mount attribute only can mean an actual read-only root, transactional system incorrectly gets identified as a non-transactional system, because read-only can also be defined as a btrfs attribute

This PR changes the is_transactional check in the snapper plugin to actually check if / is read-only, returning true if you cannot write to /

This will work regardless of the mechanism for setting root as read-only

This should help fix issues like #263 on Aeon which _only_ uses the btrfs ro attribute and not the mount/fstab ro attribute as defining both can cause kernel race conditions.

### DIFF
--- a/10-sdbootutil.snapper
+++ b/10-sdbootutil.snapper
@@ -6,7 +6,7 @@ set -e
 # check whether it's a transactional system
 is_transactional()
 {
-	findmnt --fstab / -O ro >/dev/null
+	[ ! -w /usr ] && [ -w /etc ]
 }
 
 # when creating a snapshot we fetch all bls configs from previous snapshot dir,
--- a/sdbootutil
+++ b/sdbootutil
@@ -333,7 +333,7 @@ reset_rollback()
 
 is_transactional()
 {
-	findmnt --fstab / -O ro >/dev/null
+	[ ! -w /usr ] && [ -w /etc ]
 }
 
 keyctl_add_with_timeout()


Vogtinator:
> Yeah I'm not sure what I could do about that sideeffect besides relying on zypper to do the right thing. Any suggestions?

Only `rpm -q read-only-root-fs`, but that has some overhead and relies on implementation details...

> I've changed the test to use /usr, no harm in being more future ready :)

:+1:

TobiPeterG:
Looks good to me :)
@sysrich did you test the changes on Aeon and TW?
I could do some TW testing in my VM :)

sysrich:
Yes, I've tested this

Can we merge and ship it?

TobiPeterG:
Afaik only @aplanas

aplanas:
> This will work regardless of the mechanism for setting root as read-only

Does makes sense to have this in MicroOS?

> Well I do as an org admin but I'll give @aplanas a ping in slack in the morning before considering such madness :)

I am on vacation, and with some limited access to the work laptop. The PR LGTM and I will merge it.  Can this wait until monday, to do a proper release?

There are other components that I checked, mainly suse-module-tools and perl-bootloader, but seems that this is the only component that is making this check.

sysrich:
As you're on vacation, absolutely, it can wait - but you might find a SR in OBS with me trying to do it for you when you're back :) thanks for the merge!

PR #266, opened 2025-07-16
Determine if is transactional checking read-only-root-fs

aplanas:
To determine if we are in a transactional system we can check if fstab is marking / as ro.  This works in MicroOS but not in Aeon.

The current check of non-writable /usr is also broken when the image is building, as /usr can be writable for a period of time.

Checking if the read-only-root-fs via RPM mostly works, but has the risk of interferring with the scriptlets.

A future configuration file, of a fix in the fstab marker (ro=vfs?) can provide a more robust check.

### DIFF
--- a/10-sdbootutil.snapper
+++ b/10-sdbootutil.snapper
@@ -6,7 +6,9 @@ set -e
 # check whether it's a transactional system
 is_transactional()
 {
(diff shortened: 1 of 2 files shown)


Vogtinator (approved):
Ugly but for implementation details that's ok...

PR #271, opened 2025-07-22
Measure kernel in PCR4 for grub2-bls if secure-boot

aplanas:


sdbootutil
```diff
@@ -2387,6 +2392,34 @@ pcrlock_grub2_bls()
 		done < <(jq --raw-output '.[] | .options, .linux, .initrd[0]' "$initialentryfile")
 	fi
 
+	# With secure boot, grub2-bls will make shim to extend PCR4
+	if is_secure_boot; then
```
Vogtinator, line 2396: Why not make this part of `pcrlock_grub2_bls_kernel_initrd`?

Vogtinator:
According to https://www.gnu.org/software/grub/manual/grub/html_node/UEFI-secure-boot-and-shim.html, the `MokSbState` efivar can also influence this but let's just ignore that

sdbootutil
aplanas: It is mainly about the mental model, but in this particular case will escape the policyor limitation.  For PCR4 we have a separator, so it is limited to 4 hashes, and the combination of initrd + kernel is limited to 8.

PR #273, opened 2025-07-23
measure-pcr-validator.service: Fix triggering OnFailure

Vogtinator:
- Use poweroff.target to stop other services properly (they have Conflicts
on shutdown.target).
- Run after initrd-root-device.target: local-fs.target in the initrd refers to the initrd's file system and
does not make much sense here. Use initrd-root-device.target instead.

### DIFF
--- a/measure-pcr-validator.service
+++ b/measure-pcr-validator.service
@@ -6,7 +6,7 @@ FailureAction=poweroff-immediate
 
 Wants=cryptsetup.target
 After=cryptsetup.target
-Before=local-fs.target
+Before=initrd-root-device.target
 
 [Service]
 Type=oneshot
@@ -18,5 +18,5 @@ StandardOutput=tty
 StandardInput=tty
 
 [Install]
-# If we use RequiredBy we trigger the debug shell
-WantedBy=local-fs.target
+# If we use RequiredBy we trigger its OnFailure=emergency.target
+WantedBy=initrd-root-device.target
--- a/measure-pcr-validator.service
+++ b/measure-pcr-validator.service
@@ -2,7 +2,7 @@
 Description=Validate LUKS2 devices
 DefaultDependencies=false
 
-OnFailure=systemd-halt.service
+FailureAction=poweroff-immediate
 
 Wants=cryptsetup.target
 After=cryptsetup.target


measure-pcr-validator.service
```diff
@@ -2,11 +2,12 @@
 Description=Validate LUKS2 devices
 DefaultDependencies=false
 
-OnFailure=systemd-halt.service
+OnFailure=poweroff.target
```
aplanas, line 5: Oh I wanted to fix that since long ago. Thanks!!

But I was thinking to do something similar on what is done here:
https://github.com/systemd/systemd/blob/main/units/systemd-validatefs%40.service.in#L18

But instead something like `FailureAction=halt-immediate`.  The problem with poweroff is that the message about PCR15 will not be visible to the user

aplanas (changes requested):

measure-pcr-validator.service
aplanas: is there a drawback on the systemd-validatefs approach of using "FailureAction=..."?

Vogtinator: Not sure, it's the first time I've heard of that option.

FWICT when not using any of the `-immediate` variants it's exactly equivalent to `OnFailure=` with the matching target. So `FailureAction=poweroff` makes it a bit more clear than the two lines here right now.

So the question becomes what to do on failure here. IMO `halt` is not a good state: Even the kernel doesn't do anything anymore, but the system itself is kept powered on. This can result in issues with power management.

What about leaving it `poweroff` but waiting for user input? OTOH, we want health-checker to trigger if possible.

aplanas: `-immediate` give me more confidence in this case, as any race condition or bug that I introduce will be directly avoided.

> `halt` is not a good state
>  waiting for user input

Ok for me, there is a `read` in the validator, but I see that with suspicious, TBH. A read gives the opportunity to some other service or process to be executed by systemd, maybe reaching the switch root.

Vogtinator: > -immediate give me more confidence in this case, as any race condition or bug that I introduce will be directly avoided.

Which is IMO a sign that the design of the pcr-validator is broken: Instead of complaining after unlocking, it should not allow unlocking to happen by relying on the PCR15 state in the policy

aplanas: Do you mean adding a new PolicyOR PCR15=0x00...0 OR HMAC(msg, VK) into the policy?  I can try.

Vogtinator: Yeah. IMO that's something for a later refactor though, ideally with upstream involvement (like the validatefs stuff)

aplanas (approved):

PR #277, opened 2025-07-24
measure-pcr-validator: fail if the file is missing

aplanas:


Vogtinator:
What happens if PCR 15 validation is not enabled?

aplanas:
> What happens if PCR 15 validation is not enabled?

This would require a change in the initrd, right?

Vogtinator:
FWICT there can be valid cases where the file does not exist but the module is included, including setups from before the PCR15 validation got implemented. With this change, those would refuse to boot?

aplanas:
> > I can check if PCR15 has been extended with at least one entry from systemd-cryptsetup.
> 
> That sounds fragile still. Can we instead check if any extension is requested, e.g. by checking `/etc/crypttab`?

Both are the same ... if `measure-pcr=yes` is in `crypttab` then there is a PCR15 extension for that. But sure, I can check the file instead.

Vogtinator (approved):
LGTM code-wise

aplanas:
I will test both together

PR #286, opened 2025-08-05
Add easy advanced debugging

TobiPeterG:
When the debug file exists, it will be enabled. Allows easy debugging across transactions.

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -4,6 +4,15 @@
 set -e
 shopt -s nullglob
 
+DEBUG_LOG="/var/log/sdbootutil.log"
+
+if [ -f "$DEBUG_LOG" ]; then
+	exec 3>>"$DEBUG_LOG"
+	export BASH_XTRACEFD=3
+	export PS4='+ \D{%F %T} ${BASH_SOURCE##*/}:${LINENO}:${FUNCNAME[0]:-main}: '
+	set -x
+fi
+
 verbose=
 nl=$'\n'
 shimdir="/usr/share/efi/$(uname -m)"


sdbootutil
```diff
@@ -4,6 +4,15 @@
 set -e
 shopt -s nullglob
 
+DEBUG_LOG=/boot/efi/sdbootutil-debug.log
```
aplanas, line 7: not a bad idea! I would move it somewhere to /run, maybe one file to sinalize the debug, and another to store the logs?

sdbootutil
```diff
@@ -4,6 +4,15 @@
...
+
+if [[ -f $DEBUG_LOG ]]; then
+	exec 3>>"$DEBUG_LOG"
```
aplanas, line 10: does this open a new fd (3) and funnel all the set -x commands to this 3 fd? That is cool!

sdbootutil
aplanas: Uhm no, /run is not seen from inside the transaction. IIRC it is not bindmounted by tukit. Maybe we can explore /dev/shm?  I think combustion is using this as a communication channel between inside the transaction and outside. It is very hacky.

> What would the advantage be of differenciating between the "do debug" file and the log file?

Yeah ... not much.

TobiPeterG: Ah could it be that /var/log is available inside and outside transactions? I just did a quick test and the changes done inside appeared outside :)

So should we just create the log in the directory meant for logs? :D

aplanas: Seems that indeed it is mounted:

https://github.com/openSUSE/transactional-update/blob/master/lib/Transaction.cpp#L121

That would be ideal

sdbootutil
```diff
@@ -4,6 +4,15 @@
 set -e
 shopt -s nullglob
 
+DEBUG_LOG=/var/log/sdbootutil.log
+
+if [[ -f $DEBUG_LOG ]]; then
```
aplanas, line 9: Use `[` here? And quote the variable?

aplanas (commented):
LGTM, some nits remaining

sdbootutil
aplanas: after that pass shellcheck if you can, should be all green

aplanas:
> Sure, I guess it would also be useful if users report bugs to have a issue template that asks for the output of this.

That is dangerous to do as default, as will log passwords and secrets (btw, we should add this in the wiki), but if we document it, the user can have a lot of information and they can decide what to share.

TobiPeterG:
> > Sure, I guess it would also be useful if users report bugs to have a issue template that asks for the output of this.
> 
> That is dangerous to do as default, as will log passwords and secrets (btw, we should add this in the wiki), but if we document it, the user can have a lot of information and they can decide what to share.

I added the wiki section, please take a look. :)

aplanas:
> I added the wiki section, please take a look. :)

Pretty cool, honestly. I would suggest:

* Add a command example about how to enable it (touch /var/log/...; chmod ...)
* Command can be marker with _ or some other mark
* `and watch the log:` This seem that you want to write `tail -f /var/log/...` as example?
* Thanks for adding the security consideration sections. I would add as example of leaked information: passwords and the volume key

TobiPeterG:
> What do you think of enabling this with `--debug`?

Hmm this sounds too easy honestly. Since there are security implications, this should only be done if the user is aware of them. The --debug flag might be easy to find accidentally without really knowing what happens.

PR #294, opened 2025-08-11
Support secret service to store secrets

TobiPeterG:
This currently uses secret-tool. However, it seems that this is its own password store and not using e.g. kdewallet or the GNOME keyring. I found this article https://techtea.io/articles/2025/manage-secrets-kde-wallet/ on how to query the kde wallet, but is there a tool to automatically detect the secret service and use it through a unified interface?

This however, seems to work already. An integration with the actual secret services would be nicer though.

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -33,6 +33,13 @@ arg_ask_key_pin_or_pw=
 arg_method=
 arg_signed_policy=
 arg_measure_pcr=
+
+arg_store_secrets=
+arg_secret_user=
+secret_user=
+secret_uid=
+secret_ok=
+
 have_snapshots=
 # for x in vmlinuz image vmlinux linux bzImage uImage Image zImage; do
 image=
@@ -125,6 +132,10 @@ helpandquit()
 		  --pcr			Comma seperated list of PCRs to enroll
 		  --devices		Comma separated list of devices to enroll or unenroll
 		                        By default all (not ignored) devices are [un]enrolled
+		  --store-secrets     yes|no|ask   Persist secrets in desktop keyring
+		  						Default: ask if interactive
+		  --secret-user       USER|UID     Which desktop session/user keyring to use
+		  						Default: auto-detect
 		  -v, --verbose		More verbose output
 		  -h, --help		This screen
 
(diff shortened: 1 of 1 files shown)


aplanas:
Why not use systemd-creds? The TPM2 and/or the secret host key can be used to store the passwords. They work on servers too and would not be a new dependency. Also will not require new parameters (or maybe one --persist-keys, or --no-persist-keys, depending if we opt-in or opt-out the feature)

TobiPeterG:
> Why not use systemd-creds? The TPM2 and/or the secret host key can be used to store the passwords. They work on servers too and would not be a new dependency. Also will not require new parameters (or maybe one --persist-keys, or --no-persist-keys, depending if we opt-in or opt-out the feature)

Would you recommend not using the user keyring at all? I liked that idea because that's where every other application also stores its passwords.
I also installed a VM with GNOME and secret-tool stores passwords in the GNOME keyring. It doesn't on KDE because the KDE wallet doesn't implement the secret service definition.

I had a quick look at systemd-creds, it's a nice tool :)
I wonder how we should store the credentals. Binding to the TPM would break the password store when the TPM e.g. locks, which is at least for me the primary case why I need to enter a secret in sdbootutil. Should we rely on the host key?

I would be willing to implement support for kde wallet, secret-store (already there) and if both do not exist, systemd-creds

What do you think? This way, desktop users can see their passwords in their wallet and servers don't need an extra dependency. :)

aplanas:
> Binding to the TPM would break the password store when the TPM e.g. locks

I see. If the TPM2 locks would not this indicate a security issue? If the user does not want to have TPM2 locks, why not disable it?

> I would be willing to implement support for kde wallet, secret-store (already there) and if both do not exist, systemd-creds

That is a lot of work ... I pbly will not maintain it in the future. The problem that I see is that it is currently changing the keyctl part, that I want / need to refactor.  This part is important as in an interface for other services to communicate with it.

I understand that your idea is to improve the desktop experience, that is totally ignored by my side. Maybe we can find a way to integrate this, but the current approach it very confusing for me (for example: why xtrace is required, calls to sudo, ...) I would think first in what is the experience that you want and work from there.

Maybe we can encapsulate this in some external tool that manage secrets with a single interface, similar to what hmac is doing.

aplanas:
I was checking ksecretservice and seems to be archived: https://invent.kde.org/utilities/ksecrets

Do you know if this is dead in KDE?

I am wondering if we can do something less ambitious and use `ksshaskpass6` (`/usr/libexec/ssh/ksshaskpass`) or `openssh-askpass-gnome` (`/usr/libexec/ssh/gnome-ssh-askpass`) for asking the password. The password, from the PoV of sdbootutil is always stored in the keyctl, but IIUC the tools can be used to keep the password in the desktop keyring (kwallet or gnome keyring). Does this makes sense?

TobiPeterG:
> There is a wrapper, `/usr/libexec/ssh/ssh-askpass`, but is kind of a joke. In servers it will fail, but still can be used.

Hmm but then again, we would need to support 2 stores.

aplanas:
@TobiPeterG I will close this for now.  If we work on this in the future, we can reopen it

PR #298, opened 2025-08-13
Distiguish between path and id for boot entries

danyspin97:


aplanas:
Did you tested it? Do you think that should be moved out of draft mode?

aplanas:
Ping?

sdbootutil
```diff
@@ -1159,15 +1159,16 @@ cleanup_entries()
 
 	local id snapshot kernel_version subvol src
 	while read -r id; do
+		read -r path
```
Vogtinator, line 1171: Needs to be declared as local above

danyspin97:
> Did you tested it? Do you think that should be moved out of draft mode?

Hi Alberto! I am currently investigating a jq warning to see if it is a new issue created by this PR.

PR #302, opened 2025-08-15
Don't mount /etc in chroot with btrfs subvolume

laenion:
This is an oversight from the migration to the btrfs subvolume based layout of /etc (PR #196): With the change the code would now suddenly mount /etc, and even more unfortunately that one of the current snapshot. Just skip /etc if it's identified as a btrfs mount. The code is only executed on read-only systems, so this is a valid limitation.


aplanas:
Does this fix https://bugzilla.suse.com/show_bug.cgi?id=1246043 ?

aplanas (approved):

PR #322, opened 2025-10-28
Drop systemd-experimental requirement

aplanas:


sdbootutil.spec
```diff
@@ -42,11 +42,8 @@ Requires:       qrencode
 Requires:       sed
 Requires:       (%{name}-snapper if (snapper and btrfsprogs))
 Requires:       (%{name}-tukit if read-only-root-fs)
-# While systemd-pcrlock is in experimental
-Requires:       systemd-experimental
-# something needs to require it. Can be us.
 Requires:       tpm2.0-tools
-# While bootctl is in udev
+# For bootctl and systemd-pcrlock
 Requires:       udev
```
Vogtinator: `>= 257` please, resp. whatever version the merge was done

Vogtinator: Or do `(udev >= 257 or systemd-experimental < 257)` to fix it on SLES 16

Vogtinator: Technically 16.0 has sdbootutil as well and 16.1 might not get a new systemd. Making requirements more explicit is IMO always better. Unfulfilled deps >>> silent breakage

Issue #347, opened 2026-01-15
[bug]: Adding USB stick with LUKS2 partition breaks sdbootutil


mwilck:
### Pre‑submission checklist

- [x] I searched existing issues and discussions to avoid duplicates
- [x] I tested with the latest released packages or the current git (if applicable)
- [x] I can reproduce this on a clean boot or after reboot

### Short summary

If an external disk (e.g. USB stick) with a LUKS2 partition is added to the system, sdbootutil ceases to work correctly

### Observed behavior

```text
Normal operation (no USB stick with LUKS2 partition attached):

# sdbootutil list-devices  2>/dev/null
/dev/nvme0n1p2
/dev/nvme0n1p3
/dev/nvme0n1p4
```

Now add an USB stick with a LUKS2 partition:

```
# sdbootutil list-devices  2>/dev/null
/dev/sdb3
/dev/nvme0n1p2
/dev/nvme0n1p3
/dev/nvme0n1p4
```

Note: No partition of the stick is mounted.

While this looks non-fatal in this context, it has fatal consequences for other sdbootutil subcommands. For example, consider a situation where the user has added the recovery key for `nvme0n1p2` to the `@u:cryptenroll` key in the kernel keyring. This key will most probably not work as recovery key for `sdb3`. If any `systemd-cryptenroll` command is run that needs access to the LUKS key slots, sdbootutil loops over the output of `list_devices`, where `sdb3` comes first. Because the `@u:cryptenroll` pass phrase is incorrect for the SCSI device, `systemd-cryptenroll` will ask the user for the pass phrase, which is highly unexpected.

When this happened to me, I did not realize that I was asked for the pass phrase for a "wrong" device, because I didn't ever expect any pass phrase other than that for the actually sdbootutil-managed devices to be required.

`lsblk` output:

```
/dev/sda                        
/dev/sda1           ext4        8c83e8f1-54fa-444b-9505-51ecdf0aace6
/dev/sda2           crypto_LUKS dc54e606-6b76-4dd3-a742-a1fbb5fcea26
/dev/sdb            iso9660     2023-05-11-23-47-18-61
/dev/sdb1           vfat        0AF4-0B03
/dev/sdb2           xfs         5a3ac557-8804-421d-8091-569f93d1f0ce
/dev/sdb3           crypto_LUKS 142e3c30-7f9a-4a1a-8e16-052a305f7e47
[...]

PR #348, opened 2026-01-15
sdbootutil: ignore devices that aren't listed in /etc/crypttab

mwilck:
Simply adding an external device (e.g. USB stick) with a LUKS2-encrypted
partition to a system may cause sdbootutil to fail.
E.g. systemd-cryptenroll will ask the user for a password where it
normally wouldn't.

As sdbootutil is meant to handle the devices the system needs for booting,
and expects the pass phrase for all handled devices to be the same, it
makes no sense for it to try and handle external devices. On the other hand,
It makes a lot of sense for users to add all devices that are necessary
for booting to the /etc/crypttab file. Therefore ignoring devices which
are not listen in /etc/crypttab shouldn't have negative effeccts on the system.

sdbootutil already supports excluding partitions that are marked with
`x-sdbootutil.ignore` in /etc/crypttab. But that requires that users list
every external LUKS2 device in their crypttab files, which is usually not
desired, because it has effects on system boot.

Therefore simply ignore LUKS2 devices that aren't listed in /etc/crypttab.

Fixes: https://github.com/openSUSE/sdbootutil/issues/347

(Note: this PR includes #346)

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -3085,8 +3085,11 @@ detect_tracked_devices()
 	while read -r dev fstype uuid; do
 		[ "$fstype" = 'crypto_LUKS' ] || continue
 		cryptsetup isLuks --type luks2 "$dev" || continue
-		[ -e /etc/crypttab ] && grep -E -q "${dev}[[:space:]].*x-sdbootutil.ignore" /etc/crypttab && continue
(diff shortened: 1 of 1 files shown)


aplanas:
Can you rebase? The first commit should be already merged

aplanas (approved):

PR #394, opened 2026-06-14
Add configurable devicetree entry support

gaato:
## Summary

Add `DEVICETREE_SOURCE` and `--devicetree-source` to copy a configured DTB into the boot partition and reference it from generated Boot Loader Specification Type 1 entries.

The configured source path must be absolute. It may use `%K` for the full kernel version and `%V` for the kernel package version without flavor.

Paths below the configured boot partition are resolved from the boot partition. Other absolute paths are resolved relative to the target root filesystem/snapshot.

If configured, the DTB is required: missing or unreadable files abort kernel installation instead of generating an entry that may not boot on DT-only hardware.

## Notes

This is explicit configuration, not board detection.

Currently this is limited to Secure Boot disabled. External devicetree files are not supported with Secure Boot enabled.

For `grub2-bls`, TPM prediction includes the generated `devicetree` command line in PCR 8 and the DTB file contents in PCR 9. For `systemd-boot`, TPM prediction behavior with external devicetree files has not been validated here. The Rock 5 ITX test system used below does not provide TPM2 support, so the implementation emits a warning when systemd-boot predictions are generated with `DEVICETREE_SOURCE` configured.

## Tested

- `bash -n sdbootutil`
- `bash -n completions/bash_sdbootutil`
- `shellcheck sdbootutil`
- `git diff --check origin/main...HEAD`
- `grub2-bls` prediction path updated to include `devicetree` in PCR 8 and PCR 9 prediction inputs
- Earlier OBS build in `home:gaato:rockchip/sdbootutil` for the systemd-boot devicetree path
  - `openSUSE_Tumbleweed / aarch64`: succeeded
  - `openSUSE_Factory_ARM / aarch64`: succeeded
- Earlier Rock 5 ITX systemd-boot validation running openSUSE MicroOS
  - installed `sdbootutil-1+git20260614.75efe30-9.1`
  - booted snapshot `#47`
  - current/default BLS entry: `opensuse-microos-7.0.11-1-default-47.conf`
  - generated entry contains `devicetree`
  - `sdbootutil show-entry 7.0.11-1-default 47` shows the `devicetree` field

### DIFF
--- a/sdbootutil
+++ b/sdbootutil
@@ -45,6 +45,7 @@ arg_pcr=
 arg_rootfs=
 arg_rootfs_data=
 arg_esp_free_space=
+arg_devicetree_source=
 arg_force=
 arg_disable_predictions=
 have_snapshots=
@@ -146,6 +147,15 @@ helpandquit()
 		                        (uuid, partuuid, label, partlabel, device)
 		  --rootfs-data		Extra data when --rootfs is used
 		  --esp-free-space	Percentage of free space in the ESP
+		  --devicetree-source	Absolute source path of a device tree blob to copy
+		                        into the boot partition and reference from generated
+		                        BLS entries. The source path is resolved in the
+		                        target rootfs/snapshot. %K expands to the full
+		                        kernel version and %V to the kernel package version
+		                        without the flavor suffix. If configured, the
+		                        resolved file must exist and be readable, and
+		                        must be usable without Secure Boot, or kernel
+		                        installation is aborted.
 		  --force		Force certain operations inside a transaction
 		  --disable-predictions	Disable automatic update predictions
 		  -v, --verbose		More verbose output
@@ -356,6 +366,7 @@ set_default_config_values()
 	ROOTFS="${arg_rootfs:-$ROOTFS}"
 
 	ESP_FREE_SPACE="${arg_esp_free_space:-10}"
+	DEVICETREE_SOURCE="${arg_devicetree_source:-}"
 }
 
 create_default_config_file()
@@ -376,6 +387,15 @@ create_default_config_file()
 	# Percentage (%) of free space in the ESP that sdbootutil should guarantee
 	# Default values is 10%
 	ESP_FREE_SPACE="$ESP_FREE_SPACE"
+
+	# Optional absolute path to a device tree blob. If set, it is copied to the
+	# boot partition and a devicetree field is added to generated BLS entries.
+	# The source path is resolved in the target rootfs/snapshot. %K expands to
+	# the full kernel version, %V to the kernel package version without the
+	# flavor suffix. If set, the resolved file must exist and be readable,
+	# otherwise kernel installation aborts. Currently supported only with
+	# Secure Boot disabled.
+	DEVICETREE_SOURCE="$DEVICETREE_SOURCE"
 	EOF
 }
 
@@ -410,6 +430,7 @@ load_config_file()
 	FDE_SEAL_PCR_LIST="${arg_pcr:-$FDE_SEAL_PCR_LIST}"
 	ROOTFS="${arg_rootfs:-$ROOTFS}"
 	ESP_FREE_SPACE="${arg_esp_free_space:-$ESP_FREE_SPACE}"
+	DEVICETREE_SOURCE="${arg_devicetree_source:-$DEVICETREE_SOURCE}"
 }
 
 is_secure_boot()
@@ -1028,6 +1049,16 @@ pending_initrds_size()
 	echo $((size / 1024 + 1))
 }
 
+pending_devicetree_size()
+{
+	local devicetree="$1"
+	[ -n "$devicetree" ] || {
+		echo 0
+		return 0
+	}
+	echo $(($(stat -L -c %s "$devicetree") / 1024 + 1))
+}
+
 boot_space()
 {
 	echo $(($(findmnt --noheadings --bytes -o SIZE --target "${boot_root}" | head -n 1) / 1024))
@@ -1178,13 +1209,33 @@ make_free_space_for_kernel()
 	local snapshot="$1"
 	local kernel="$2"
 	local initrds="$3"
+	local devicetree="${4:-}"
 
 	# Calculate the free space and the required size.  All sizes
 	# are in Kb to avoid big numbers
 	local free_space total_size
-	total_size=$(($(pending_kernel_size "$kernel") + $(pending_initrds_size "$initrds")))
+	total_size=$(($(pending_kernel_size "$kernel") + $(pending_initrds_size "$initrds") + $(pending_devicetree_size "$devicetree")))
 
 	make_free_space "$snapshot" "$total_size"
+}
+
+resolve_devicetree_source()
+{
+	local subvol="$1"
+	local kernel_version="$2"
+	local kernel_package_version="${kernel_version%-*}"
+	local devicetree_source="${DEVICETREE_SOURCE:-}"
+
+	[ -n "$devicetree_source" ] || return 0
+	devicetree_source="${devicetree_source//%K/$kernel_version}"
+	devicetree_source="${devicetree_source//%V/$kernel_package_version}"
+	[[ "$devicetree_source" = /* ]] || err "DEVICETREE_SOURCE must be an absolute path"
+
+	local snapshot_prefix="${subvol#"${subvol_prefix}"}"
+	devicetree_source="$snapshot_prefix$devicetree_source"
+	[ -f "$devicetree_source" ] || err "Can't find device tree blob $devicetree_source"
+	[ -r "$devicetree_source" ] || err "Device tree blob is not readable: $devicetree_source"
+	echo "$devicetree_source"
 }
 
 create_boot_options() {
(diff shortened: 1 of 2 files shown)


sdbootutil
```diff
@@ -3456,6 +3537,13 @@ generate_tpm2_predictions()
 		return 0
 	}
 
+	[ -z "${DEVICETREE_SOURCE:-}" ] || ! is_grub2_bls || {
```
aplanas: Worried about this.  Is the kernel the one that measure it? Is doing a simple `systemd-pcrlock raw`? I think that we can implement the measurement as part of this PR

gaato: Thanks. I added grub2-bls devicetree prediction support.

The generated `devicetree` command line is included in PCR 8 prediction, and the DTB file contents are included in PCR 9 prediction.

sdbootutil
```diff
@@ -1244,9 +1304,21 @@ install_kernel()
...
+	local devicetree_dst=
 
 	local initrd="${src%/*}/initrd"
 
+	# Resolve the device tree before creating any ESP directory, so a
+	# missing (but configured) blob aborts without leaving an empty
+	# "$entry_token/$kernel_version" directory behind
+	devicetree_src="$(resolve_devicetree_source "$subvol" "$kernel_version")" || err "Failed to resolve DEVICETREE_SOURCE"
+	if [ -n "$devicetree_src" ]; then
+		! is_secure_boot || err "External devicetree files are not supported when Secure Boot is enabled"
```
aplanas, line 1310: Can they be signed?

sdbootutil
```diff
@@ -3456,6 +3560,10 @@ generate_tpm2_predictions()
 		return 0
 	}
 
+	[ -z "${DEVICETREE_SOURCE:-}" ] || ! is_sdboot || {
```
aplanas, line 3559: Thanks for adding the prediction for GRUB2, we should add the one for systemd-boot too if possible.

When systemd-stub (UKI) is used then PCR12 is extended. We need PCR15 that is also extended only when systemd-stub is used, that is why we add this override to systemd-cryptsetup (via a generator)

https://github.com/openSUSE/sdbootutil/blob/main/measure-pcr-generator.sh#L28

I guess that this is useless in this case, and the warning here makes sense.  Did you check that the devicetree entry is not measured by systemd-boot in any case?

sdbootutil
gaato: I don't think they can be signed in this external BLS `devicetree` path. Since systemd-boot skips external devicetree files when secure boot is enabled, this PR rejects that combination.

https://github.com/systemd/systemd/blob/main/src/boot/boot.c#L3015-L3021

sdbootutil
gaato: I checked the systemd-boot path and could not find a measurement of the external `devicetree` file itself. `devicetree_install()` reads the DTB, applies fixups, and installs it into the EFI configuration table, but I do not see a TPM measurement call in that path:

https://github.com/systemd/systemd/blob/main/src/boot/devicetree.c#L65-L106

So I kept this as a warning rather than adding a prediction that may not match the event log.

sdbootutil
```diff
@@ -1176,15 +1208,43 @@ make_free_space()
 make_free_space_for_kernel()
 {
 	local snapshot="$1"
+	local devicetree="${2:-}"
```
aplanas: I see. I like this change, and I updated master to pass also the kernel and initrds path.  This would require a rebase, sorry

sdbootutil
```diff
@@ -146,6 +147,16 @@ helpandquit()
 		                        (uuid, partuuid, label, partlabel, device)
 		  --rootfs-data		Extra data when --rootfs is used
 		  --esp-free-space	Percentage of free space in the ESP
+		  --devicetree-source	Absolute source path of a device tree blob to copy
+		                        into the boot partition and reference from generated
+		                        BLS entries. Paths below the boot partition are
+		                        resolved there; other paths are resolved in the
+		                        target rootfs/snapshot. %K expands to the full
+		                        kernel version and %V to the kernel package version
```
aplanas, line 154: What is the difference between kernel version and kernel package version?

sdbootutil
```diff
@@ -1176,15 +1208,43 @@ make_free_space()
...
+	devicetree_source="${devicetree_source//%V/$kernel_package_version}"
+	[[ "$devicetree_source" = /* ]] || err "DEVICETREE_SOURCE must be an absolute path"
+
+	local snapshot_prefix="${subvol#"${subvol_prefix}"}"
+	local boot_root_path="${boot_root%/}"
+	[ -n "$boot_root_path" ] || boot_root_path="/"
+
+	if [ -n "$snapshot_prefix" ] && [ "$boot_root_path" != "/" ]; then
+		case "$devicetree_source" in
+			"$boot_root_path"|"$boot_root_path"/*) ;;
```
aplanas: I am not sure to understand all this changes, why the final place for the device tree file is not `"${subvol#"${subvol_prefix}"}/${devicetree_source}"`

sdbootutil
aplanas: Sorry for my ignorance, but if the dtb is loaded by the kernel, then is not measured by the kernel itself? It should be a similar case with initrd; with systemd-boot it is the kernel the one that loads it and extend the corresponding PCR.

Are you testing this with QEMU with a TPM2?

sdbootutil
```diff
@@ -1312,6 +1384,7 @@ install_kernel()
 	options    $boot_options
 	linux      $dst
 	EOF
+	[ -z "$devicetree_dst" ] || echo "devicetree $devicetree_dst" >> "$tmpdir/entry.conf"
```
aplanas: Do you mind to move it into the  `cat <<-EOF` block? There is an example of an optional field for the `machine-id` field

sdbootutil
```diff
@@ -1324,6 +1397,13 @@ install_kernel()
 	else
 		info "Reusing ${boot_root}$dst"
 	fi
+	if [ -z "$failed" ] && [ -n "$devicetree_src" ]; then
+		if [ ! -e "${boot_root}$devicetree_dst" ]; then
```
aplanas: nit: ${devicetree_dst} when expanding variables with _ (I am not following this rule everywhere, but I try)

sdbootutil
```diff
@@ -2642,8 +2722,10 @@ pcrlock_grub2_bls_kernel_initrd_cmdline_initrd()
 	local initrd="$2"
 	local cmdline="$3"
 	local suffix="$4"
+	local devicetree="${5:-}"
```
aplanas: IMHO this should be after the initrd parameter.  I think you can send empty parameters in bash:

```bash
a() {
  x="$1"; y="$2"; z="$3"
  echo "x: $x, y: $y, z: $z"
}

a "1" "2" "3"
a "1" "" "3"
```

aplanas (commented):
Some comments, but basically LGTM

sdbootutil
gaato: `%K` expands to the full kernel version including the flavor suffix, for example `7.0.11-1-default`.

`%V` expands to the kernel package version without the flavor suffix, for example `7.0.11-1`. I clarified the wording to say "without the flavor suffix".

sdbootutil
gaato: That path is only used to resolve the source DTB in the target rootfs/snapshot.

The installed DTB is intentionally copied into the boot partition next to the copied kernel and initrd, under `$BOOT/<entry-token>/<kernel-version>/`. The generated BLS `devicetree` field then points to that boot-partition path, so the boot loader can read it and `bootctl cleanup` can manage it with the rest of the generated entry files.

sdbootutil
gaato: Done.

sdbootutil
gaato: Done.

sdbootutil
gaato: Done. I moved the `devicetree` argument after `initrd` and pass an explicit empty argument when no devicetree is present.

sdbootutil
gaato: I tested this with an aarch64 QEMU/libvirt VM using swtpm/TPM2 and systemd-boot 260.2.

I compared Type 1 entries generated by this PR's `sdbootutil` with and without an external `devicetree` line. The generated entries differed only by that line, but the binary TPM event logs and PCR 9/12/15 were identical.

I could not find evidence that the external DTB is measured in the systemd-boot path. That is why I left the systemd-boot prediction path unchanged.

aplanas: Thanks!

sdbootutil
aplanas: I think that I do not like that it is OK for the device tree source to be the ESP, in the same sense that is not OK for the kernel. I would expect something like this (not tested):

```bash
resolve_devicetree_source()
{
	local subvol="$1"
	local kernel_version="$2"
	local kernel_package_version="${kernel_version%-*}"
	local devicetree_source="${DEVICETREE_SOURCE:-}"

	[ -n "$devicetree_source" ] || return 0
	devicetree_source="${devicetree_source//%K/$kernel_version}"
	devicetree_source="${devicetree_source//%V/$kernel_package_version}"
	[[ "$devicetree_source" = /* ]] || err "DEVICETREE_SOURCE must be an absolute path"

	devicetree_source="${subvol#"${subvol_prefix}"}/${devicetree_source}"
	[ -f "$devicetree_source" ] || err "Can't find device tree blob $devicetree_source"
	[ -r "$devicetree_source" ] || err "Device tree blob is not readable: $devicetree_source"
	echo "$devicetree_source"
}
```

Am I missing something?

aplanas:
Can you squash the commits? I would like to merge them as a single change

sdbootutil
gaato: No, that makes sense.

I simplified this so `DEVICETREE_SOURCE` is always resolved relative to the target rootfs/snapshot, like the kernel source. The resolved DTB is still copied into the boot partition next to the generated kernel/initrd files.

aplanas (approved):
Thanks! This feature is very useful

